PR #25, opened 2024-07-10
fix(FunctionDetails): Get timeline state from Flame Graph component

bryanhuhta:
### ✨ Description

related: #15 

### 📖 Summary of the changes

In #15, I added some code to sync  the time range with the query client for function details. This approach was flawed (https://github.com/grafana/explore-profiles/pull/15#discussion_r1671943277).

This PR moves logic of setting the function details time range to the Flame Graph scene component. There's also some minor clean up to properly handle the error case when the time range that powers the function details component is not set. This should never happen, but if it does, the app won't crash. Instead, we'll get an error message like this:

Screenshot 2024-07-10 at 11 53 25 AM

### 🧪 How to test?

- tests pass
- Run the UI and change the time range, it subsequent `SelectMergeProfile` requests should use the correct time range


.github/workflows/publish-technical-documentation-next.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     paths:
-      - "docs/sources/**"
```
grafakus, line 8: We should keep this PR focused and not include the changes in the Markdown files. I've opened https://github.com/grafana/explore-profiles/pull/59 to take care of it.

src/shared/components/FunctionDetailsPanel/infrastructure/useFetchFunctionsDetails.ts
```diff
@@ -35,15 +35,33 @@ export function useFetchFunctionsDetails({
...
+  const {
+    isFetching,
+    error: queryError,
+    data,
+  } = useQuery({
     enabled: Boolean(profileMetricId && labelsSelector && stacktrace.length > 0),
     // eslint-disable-next-line @tanstack/query/exhaustive-deps
     queryKey: ['function-details', profileMetricId, labelsSelector, start, end, stacktrace, isGitHubLogged],
     queryFn: async () => {
+      if (timeRangeError) {
```
grafakus, line 61: This should be added to the `enabled` option and not here (see [my comment](https://github.com/grafana/explore-profiles/pull/25/files#r1673652705))

src/pages/ProfilesExplorerView/exploration-types/SceneServiceFlameGraph/SceneFlameGraph.tsx
```diff
@@ -33,6 +35,27 @@ export class SceneFlameGraph extends SceneObjectBase<SceneFlameGraphState> {
...
       }),
     });
+
+    this.addActivationHandler(this.onActivate.bind(this));
+  }
+
+  onActivate() {
+    let timeRangeSubscription: Unsubscribable | null = null;
+
+    const stateSubscription = this.subscribeToState((newState) => {
```
grafakus, line 45: There are a couple of issues with this implementation:
- `$data` always exists (it's set when calling `super` L33)
- every time the state of this scene object will change, a new subscription will be created (L47) - we only need a single one
- there is still a conceptual problem: do we know under which circumstances `data.timeRange` changes? is it when starting to load the data? after loading? what we really need is to store the last timerange **that was used to successfully load the data**

My recommendation would be to use this implementation:

```js
    let dataSub: undefined | Unsubscribable;

    const stateSub = this.subscribeToState((newState, prevState) => {
      if (newState.$data !== prevState.$data) {
        if (dataSub) {
          dataSub.unsubscribe();
        }

        dataSub = newState.$data?.subscribeToState((newState) => {
          if (newState.data?.state === LoadingState.Done) {
            timelineAndProfileApiClient.setLastTimeRange(newState.data.timeRange);
          }
        });
      }
    });

    return () => {
      stateSub.unsubscribe();
      dataSub?.unsubscribe();
    };
```

.github/workflows/publish-technical-documentation-next.yml
bryanhuhta: I must have ran a linter at some point. I don't recall changing this file 😬 reverted in https://github.com/grafana/explore-profiles/pull/25/commits/4bffbdf3848d5a3197122fb1150ec86e80173a75

src/shared/components/FunctionDetailsPanel/infrastructure/useFetchFunctionsDetails.ts
bryanhuhta: My bad! fixed in https://github.com/grafana/explore-profiles/pull/25/commits/c397c75b7c37252245a90b244d3c16ca1a84ca15

src/pages/ProfilesExplorerView/exploration-types/SceneServiceFlameGraph/SceneFlameGraph.tsx
bryanhuhta: Ah gotcha, this makes sense. Fixed in https://github.com/grafana/explore-profiles/pull/25/commits/46cfccbb443190e026ceec9a16e4e9f19c4a5990

src/shared/components/FunctionDetailsPanel/infrastructure/useFetchFunctionsDetails.ts
```diff
@@ -35,12 +35,26 @@ export function useFetchFunctionsDetails({
...
+  } catch (e) {
+    timeRangeError = e as Error;
+  }
+
+  const {
+    isFetching,
+    error: queryError,
+    data,
+  } = useQuery({
+    enabled: Boolean(profileMetricId && labelsSelector && stacktrace.length > 0 && !timeRangeError),
```
grafakus, line 57:
```suggestion
    enabled: Boolean(profileMetricId && labelsSelector && stacktrace.length > 0 && start > 0 && end > 0 && !timeRangeError),
```

grafakus (approved):

PR #58, opened 2024-07-16
fix(CompareAction): Add missing data source query parameter to compare URL

grafakus:
### ✨ Description

**Related issue(s):** `-`

Before this PR, when going from the "Labels" exploration type to the compare page, the data source query parameter was not passed. If there were multiple Pyroscope data sources, it could lead to a broken UI (e.g.):

image

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- In an instance with multiple Pyroscope data sources configured
- Go to the "Labels" exploration type
- Select 2 label values for comparison
- Click "Compare" to open the "Comparison Diff View"
- The UI should always display correctly all the data, regardless of the data source that was selected


simonswine (approved):
LGTM

PR #64, opened 2024-07-17
feat(Analytics): Track Explore Profiles user actions

grafakus:
### ✨ Description

**Related issue(s):** `-`

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- Once merged, we'll have to update our internal dashboard and see the new actions being tracked.

### DIFF
--- a/src/pages/ProfilesExplorerView/variables/ProfileMetricVariable.tsx
+++ b/src/pages/ProfilesExplorerView/variables/ProfileMetricVariable.tsx
@@ -2,6 +2,7 @@ import { css } from '@emotion/css';
 import { GrafanaTheme2, VariableRefresh } from '@grafana/data';
 import { MultiValueVariable, QueryVariable, SceneComponentProps, VariableValueOption } from '@grafana/scenes';
 import { Cascader, CascaderOption, Icon, Tooltip, useStyles2 } from '@grafana/ui';
+import { reportInteraction } from '@shared/domain/reportInteraction';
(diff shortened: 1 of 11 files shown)


src/pages/ProfilesExplorerView/variables/ProfileMetricVariable.tsx
```diff
@@ -21,7 +22,7 @@ export class ProfileMetricVariable extends QueryVariable {
   constructor() {
     super({
       name: 'profileMetricId',
-      label: '🔥 Profile',
+      label: '🔥 Profile type',
```
grafakus, line 25: To be consistent with the rest of the UI (not strictly related to this PR).

src/shared/domain/reportInteraction.ts
```diff
@@ -1,10 +1,28 @@
...
+  const extraProperties: Record<string, any> = { page };
+
+  if (page === PROFILES_EXPLORER_PAGE_NAME) {
+    extraProperties.explorationType = new URLSearchParams(window.location.search).get('explorationType');
+  }
+
+  return extraProperties;
+}
+
 export function reportInteraction(interactionName: string, properties?: Record<string, unknown>) {
```
bryanhuhta, line 23: I think at some point we might want to make `interactionName` a union type of strings literals (or something similar) so we don't need to search the code to get a list of all the interaction names we have. It would be in one list here.

Something like:

```ts
type Interaction =
  'g_pyroscope_app_service_name_selected' |
  'g_pyroscope_app_profile_metric_selected' |
  ...
```

bryanhuhta (approved):
LGTM :shipit:

PR #92, opened 2024-07-25
[experiment] feat(Labels): New "Single" layout when a groupBy value is selected

grafakus:
### ✨ Description

**Related issue(s):** `-`

https://github.com/user-attachments/assets/e7ba0175-ea6a-41de-804d-c3fe5ae3575b


grafakus:
Closing, we shouldn't add a new layout but use the main timeseries to show the current label values

PR #94, opened 2024-07-26
feat(*): Revamp exploration type selector

grafakus:
### ✨ Description

**Related issue(s):** `-`

**Motivation:** new users are confused about where they are and how they can navigate back (the browser back button doesn't work). This PR changes the styles of the exploration type selector to make this more obvious. Note that, this is more a "low hanging fruit" than a real improvement. We will address the fundamental issue once our UX research is more advanced.

| Before | After |
|  ---   |  ---  |
| image | image |

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The smoke E2E tests should pass

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/ui/ExplorationTypeSelector.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/ui/ExplorationTypeSelector.tsx
@@ -30,6 +30,7 @@ export function ExplorationTypeSelector({ options, value, onChange }: Exploratio
                 onClick={isActive ? noOp : () => onChange(option.value as string)}
                 tooltip={option.description}
                 tooltipPlacement="top"
+                data-testid={isActive ? 'is-active' : undefined}
               >
                 {option.label}
               </Button>
--- a/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/ui/ExplorationTypeSelector.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/ui/ExplorationTypeSelector.tsx
@@ -1,46 +1,64 @@
 import { css } from '@emotion/css';
-import { SelectableValue } from '@grafana/data';
-import { InlineLabel, RadioButtonGroup, Select, useStyles2 } from '@grafana/ui';
+import { GrafanaTheme2 } from '@grafana/data';
+import { Button, Icon, InlineLabel, useStyles2 } from '@grafana/ui';
 import React from 'react';
 
 export type ExplorationTypeSelectorProps = {
-  layout: 'radio' | 'select';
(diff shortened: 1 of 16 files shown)


Rperry2174:
I like this! I feel like this is the most important thing we want people to get from the app at first glance...

What if we make this on the top "row" by itself and then beneath we have datasource and service name next to each other?

grafakus:
> What if we make this on the top "row" by itself and then beneath we have datasource and service name next to each other?

The issue is that this part of the UI will start to feel a bit crowded:
image

Also, once a data source is selected, the users will most likely play with the service and profile type selectors, so it makes sense to separate the data source selector on the UI IMO.

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/ui/ExplorationTypeSelector.tsx
```diff
@@ -1,52 +1,71 @@
...
   onChange: (newValue: string) => void;
 };
 
-export function ExplorationTypeSelector({ layout, options, value, onChange }: ExplorationTypeSelectorProps) {
+export function ExplorationTypeSelector({ options, value, onChange }: ExplorationTypeSelectorProps) {
   const styles = useStyles2(getStyles);
 
   return (
     <div className={styles.explorationTypeContainer} data-testid="exploration-types">
-      <InlineLabel width="auto">Exploration type</InlineLabel>
```
simonswine, line 18: Nit: Not really happy about this world. "Exploration" is not telling me anything. Maybe "Steps" is better?

Please do not block on this, I think we can always change that word

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/ui/ExplorationTypeSelector.tsx
```diff
@@ -1,52 +1,71 @@
...
+  button: css`
+    height: 30px;
+    line-height: 30px;
+  `,
+  active: css`
+    &:hover {
+      cursor: default;
+    }
   `,
 });
```
simonswine, line 71: I guess the "main" breadcrumb is quite different in style:

image

I think I see that the button help to make it clearer where one is in the way to the flamegraph, maybe it makes sense to be closer to the main one by using `>` instead of arrows.

simonswine (approved):
I really like this way this explains the flow. 

LGTM

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/ui/ExplorationTypeSelector.tsx
grafakus: Thanks for the suggestion, we'll continue to iterate on it in the future PRs.

Issue #95, opened 2024-07-26
Label breakdown not helpful, when there's only one label value.


simonswine:
I think it might make sense to filter the label names with one label value all together in the by label breakdowns:

![image](https://github.com/user-attachments/assets/94e1527d-5661-47b4-846e-c7f98a0faa42)

There is really not much to action on when there is only a single value.

PR #108, opened 2024-08-01
feat(Timeseries): Add total resource consumption in legend

grafakus:
### ✨ Description

**Related issue(s):** `-`

Whenever the UI displays a single timeseries, the total resource consumption is displayed in the legend:
image

This PR also does some refactorings to prepare the integration of the Comparison view.

### 📖 Summary of the changes

See diff tab for specific comments

### 🧪 How to test?

`-`


simonswine:
My only comment would be, I find the incosistency weird with the single value labels (again #95)

image

simonswine (approved):
LGTM

grafakus:
> My only comment would be, I find the incosistency weird with the single value labels (again #95)

I understand, it's a temporary state to provide a different view that I demoed to @bryanhuhta yesterday. Also, adding the total for each timeseries makes things unreadable:
image

PR #110, opened 2024-08-05
fix(Labels): Fix "Discarded by user" error in the UI

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR fixes an error that can occur when changing the data source or clicking on the main refresh button in the UI:

image

image

Additionally, this PR also fixes a potential issue with the query builder, which does not update its API client whenever the data source changes.

### 📖 Summary of the changes

The error comes from an unwanted interaction between the query builder and `LabelsDataSource` where the API request made by `LabelsDataSource` to fetch label names is cancelled by the query builder state machine. This happens because the both share the same instance of the `LabelsRepository` class, which in turn uses the same API client.

The scenario is the following:

Whenever the user changes the data source:
1. `GroupByVariable` gets its options updated via `LabelsDataSource`
2. In `LabelsDataSource`, `LabelsRepository` sets the API client corresponding to the new data source
3. the query builder state machine receives new input parameters, which in turn, causes the machine to transition to its [idle state](https://github.com/grafana/explore-profiles/blob/main/src/shared/components/QueryBuilder/domain/states/idle.ts#L15)
4. when entering this state, all in-flight requests are cancelled (see [QueryBuilder/domain/actions.ts](https://github.com/grafana/explore-profiles/blob/main/src/shared/components/QueryBuilder/domain/actions.ts#L44-L46))
5. because the same API client is used by `LabelsRepository`, this throws an abort error that bubbles up to the "Group by" UI

The solution is to make sure distinct instances of the API client are created by the query builder and by `LabelsDataSource`.

See the diff tab for additional specific comments.

### 🧪 How to test?

- Manually, after checking out this PR's branch


src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/components/SceneErrorState/SceneErrorState.tsx
```diff
@@ -10,7 +10,7 @@ export class SceneErrorState extends SceneObjectBase<SceneErrorStateState> {
   public static Component = ({ model }: SceneComponentProps<SceneErrorState>) => {
     const { message } = model.useState();
     return (
-      <Alert title="Query error" severity="error">
+      <Alert title="Query error!" severity="error">
```
grafakus, line 13: Cosmetics...

src/pages/ProfilesExplorerView/domain/variables/GroupByVariable/GroupByVariable.tsx
```diff
@@ -93,16 +93,14 @@ export class GroupByVariable extends QueryVariable {
...
-          <Field label="Group by labels">
-            <div className={styles.groupByErrorContainer}>
-              <Tooltip theme="error" content={error.toString()}>
-                <Icon className={styles.iconError} name="exclamation-triangle" size="xl" />
-              </Tooltip>
-              <RefreshPicker noIntervalPicker onRefresh={model.update} isOnCanvas={false} onIntervalChanged={noOp} />
-            </div>
-          </Field>
-        </>
+        <Field label="Group by labels">
```
grafakus, line 96: Cosmetics as well

src/shared/components/QueryBuilder/domain/helpers/filtersToQuery.ts
```diff
@@ -6,9 +6,9 @@ export function filtersToQuery(query: string, filters: Filters) {
     .map((filter) => {
       const { attribute, operator, value } = filter as CompleteFilter;
 
-      if (operator.value === OperatorKind.in) {
-        return `${attribute.value}=~"${value.value}"`;
-      }
+      // if (operator.value === OperatorKind.in) {
```
grafakus, line 9: Not strictly related to this PR, but if we execute this code, the "in" operator will be rendered in the filters. This cannot be because [it's not included in the operators dropdown options](https://github.com/grafana/explore-profiles/blob/main/src/shared/components/QueryBuilder/infrastructure/operatorsRepository.ts#L12).

src/shared/components/QueryBuilder/domain/helpers/queryToFilters.ts
```diff
@@ -48,23 +48,23 @@ export function queryToFilters(query: string): Filters {
...
-          value: {
-            value: value,
-            label: value
-              .split('|')
-              .map((v) => v.trim())
-              .join(', '),
-          },
-        };
-      }
+      // const shouldChangeToInOperator = operator === OperatorKind['=~'] && value.includes('|');
```
grafakus, line 51: Same comment as above

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
```diff
@@ -9,37 +9,30 @@ import {
...
 import { GroupByVariable } from '../../domain/variables/GroupByVariable/GroupByVariable';
 import { computeRoundedTimeRange } from '../../helpers/computeRoundedTimeRange';
 import { PYROSCOPE_LABELS_DATA_SOURCE } from '../pyroscope-data-sources';
 import { DataSourceProxyClientBuilder } from '../series/http/DataSourceProxyClientBuilder';
 import { LabelsApiClient } from './http/LabelsApiClient';
 
+// we instantiate a new repository here to prevent unwanted interaction with the QueryBuilder
+// indeed, when entering the "idle" state, the QueryBuilder state machine will call cancelAllLoad()
+// that will abort any ApiClient in-flight request
+const labelsRepository = new LabelsRepository({
```
grafakus, line 24: The fix

src/shared/components/QueryBuilder/domain/helpers/queryToFilters.ts
grafakus, line 51: Not strictly related to this PR, but if we let this code being executed, the "in" operator will be rendered in the filters in the UI. This cannot be because [it's not included in the operators dropdown options](https://github.com/grafana/explore-profiles/blob/main/src/shared/components/QueryBuilder/infrastructure/operatorsRepository.ts#L12).

src/shared/components/QueryBuilder/domain/helpers/__tests__/filtersToQuery.spec.ts
```diff
@@ -60,23 +60,23 @@ const spanNameFilter = {
   },
 };
 
-const podIdFilter = {
```
grafakus, line 63: See comments below

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
```diff
@@ -9,37 +9,30 @@ import {
...
+  apiClient: new LabelsApiClient({ dataSourceUid: '' }),
```
simonswine, line 25: Why is the uid forced to be empty here, compared to https://github.com/grafana/explore-profiles/blob/4ee541acbe822d92abfc9344eda4611600b1476e/src/shared/infrastructure/labels/labelsRepository.ts#L115-L118

src/shared/components/QueryBuilder/domain/helpers/queryToFilters.ts
simonswine: IMHO We should always remove the commented code or at least make it very clear with an extra comment why it is there and when this code comes back.

simonswine (commented):
Almost LGTM

A couple of smaller questions/suggestion

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
```diff
@@ -9,37 +9,30 @@ import {
...
+  cacheClient: new MemoryCacheClient(),
```
simonswine, line 26: Is it worth reusing the cache between the two instances of this?
 https://github.com/grafana/explore-profiles/blob/4ee541acbe822d92abfc9344eda4611600b1476e/src/shared/infrastructure/labels/labelsRepository.ts#L115-L118

src/shared/components/QueryBuilder/domain/helpers/queryToFilters.ts
grafakus: Indeed, done!

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
grafakus: Because they are different class of API clients. This one, used by `LabelsDataSource` inherits from `DataSourceProxyClient`, which requires a data source UID when instantiated. This allows us to switch the data source in runtime when the user selects different data source in the UI.

The other one inherits from `ApiClient`, which determines the data source UID only when the user lands on the page for the 1st time. It's used by the (legacy) Comparison page where we don't need to switch the data source in runtime.

It's confusing, but it's temporal. Once we migrate the Comparison page, we'll only need the client inheriting from `DataSourceProxyClient`.

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
grafakus: Indeed. By making the change, I actually realised that this fix is not the best we can do. I've just commited the new fix.

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
```diff
@@ -82,22 +87,17 @@ export class LabelsDataSource extends RuntimeDataSource {
       return [];
     }
 
-    // we could interpolate ad hoc filters, but the Service labels exploration type would reload all labels each time they are modified
-    // const filters = sceneGraph.interpolate(sceneObject, '$filters');
-    // const pyroscopeQuery = `${profileMetricId}{service_name="${serviceName}",${filters}}`;
-    const pyroscopeQuery = `${profileMetricId}{service_name="${serviceName}"}`;
-
-    const { from, to } = computeRoundedTimeRange(range as TimeRange);
+    labelsRepository.setApiClient(new LabelsApiClient({ dataSourceUid }));
```
grafakus, line 90: The fix. We don't cache the API client anymore with `DataSourceProxyClientBuilder`, so that when the query builder aborts a request, it will do it via its own API client.

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
```diff
@@ -60,18 +43,40 @@ export class LabelsDataSource extends RuntimeDataSource {
     };
   }
 
-  async metricFindQuery(query: string, options: LegacyMetricFindQueryOptions): Promise<MetricFindValue[]> {
+  getParams(options: LegacyMetricFindQueryOptions) {
```
grafakus, line 46: Some refactoring.

src/shared/components/QueryBuilder/domain/actions.ts
```diff
@@ -239,12 +240,20 @@ export const actions: any = {
...
-    inputParams: (context: QueryBuilderContext, event: ChangeInputParamsEvent) => event.data,
-    query: (context: QueryBuilderContext, event: ChangeInputParamsEvent) => event.data.query,
-    // See also buildStateMachine() in domain/stateMachine.ts
-    filters: (context: QueryBuilderContext, event: ChangeInputParamsEvent) => queryToFilters(event.data.query),
-    isQueryUpToDate: true,
+  changeInputParams: assign((context: QueryBuilderContext, event: ChangeInputParamsEvent) => {
+    // TODO: remove this condition after migrating the legacy comparison pages to Scenes
+    // because dataSourceUid will always be provided
+    if (event.data.dataSourceUid) {
+      labelsRepository.setApiClient(new LabelsApiClient({ dataSourceUid: event.data.dataSourceUid }));
```
grafakus, line 247: The 2nd fix

src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
simonswine: Thanks for explaining, that makes sense 👍

simonswine (approved):
LGTM

Thanks for helping me understand this. I think this is ready to go.

PR #117, opened 2024-08-06
feat(Labels): Improve comparison flow

grafakus:
### ✨ Description

**Related issue(s):** is blocked by https://github.com/grafana/explore-profiles/pull/120

This PR aims at improving the comparison flow by adding clear CTAs in the "Labels" view. When a label has been selected:

1. There's a new "Compare" button above the grid
2. Each grid panel has a "Baseline" and a "Comparison" button, to choose the target of the comparison

Once both "Baseline" & "Comparison" have been selected, the "Compare" button becomes active:

image

Clicking on the "Compare" button still opens a new tab with the (legacy) "Comparison Diff" view. This behaviour will be improved in https://github.com/grafana/explore-profiles/pull/119.

### 📖 Summary of the changes

### 🧪 How to test?

- Manually, after checking this PR's branch locally


src/pages/ProfilesExplorerView/components/SceneExploreServiceLabels/SceneExploreServiceLabels.tsx
```diff
@@ -94,27 +89,6 @@ export class SceneExploreServiceLabels extends SceneObjectBase<SceneExploreServi
       const filtersVariable = findSceneObjectByClass(this, FiltersVariable) as FiltersVariable;
       filtersVariable.setState({ filters });
     }
-
```
grafakus, line 97: Move closer to where it's used, in `SceneGroupByLabels`

src/pages/ProfilesExplorerView/components/SceneExploreServiceLabels/components/SceneGroupByLabels/components/SceneLabelValuesGrid/SceneLabelValuesGrid.tsx
```diff
@@ -0,0 +1,424 @@
...
+    startColorIndex: SceneLabelValuesGridState['startColorIndex'];
+    headerActions: SceneLabelValuesGridState['headerActions'];
+  }) {
+    super({
+      key,
+      label,
+      startColorIndex,
+      items: [],
+      isLoading: true,
+      $data: new SceneDataTransformer({
```
grafakus, line 86: This is one of the main difference with `SceneByVariableRepeaterGrid`. Using a query with the `groupBy` parameter here improves the consistency between the grid that is displayed when "All" is selected vs when a specific label has been selected. It ensures that:
- A timeseries with an empty label value is also rendered
- The panels will be sorted by decreasing total of resource consumption (thanks to the `sortSeries` data transformation)

For example, with the `vehicle` label:

image
(notice the "cpu" timeseries which corresponds to an empty label value)

And after selection, the "cpu" timeseries also appears in the grid:

image

src/pages/ProfilesExplorerView/components/SceneExploreServiceLabels/components/SceneGroupByLabels/components/SceneLabelValuesGrid/domain/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum CompareTarget {
+  BASELINE = 'baseline',
+  COMPARISON = 'comparison',
```
simonswine, line 3:
```suggestion
  BASELINE = 'baseline',
  COMPARISON = 'comparison',
```

The behaviour of only have either or is quite limiting. Assuming you have 3 pods a,b,c

I think it is very valid and a good usecase to compare:

baseline: b,c (or everything not a)
comparsion: a

Even this makes a lot of sense

baseline: b,c (or everything not a)
comparsion: a

Wonder if that is something you were planning to address later (once you have the "in" support)

simonswine (dismissed):
The flow is a lot clearer this way, I still would love to see a comparison right once you select either baseline and comparision in the top time line.

Also limit to one selection each is preventing a lot of important flows, but that might not be something you were planning to do in this PR. 

LGTM

src/pages/ProfilesExplorerView/infrastructure/helpers/getSeriesLabelFieldName.ts
```diff
@@ -1,4 +1,4 @@
 import { Field } from '@grafana/data';
 
 export const getSeriesLabelFieldName = (metricField: Field, label?: string) =>
-  metricField.labels?.[label as string] || metricField.name;
+  metricField.labels?.[label as string] || metricField.name; // metricField.labels can be empty when the ingested profiles do not have a label value set
```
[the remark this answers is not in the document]
grafakus: This is Grafana's default behaviour. 

Note that I used "?" in the past but it was not understood. An empty value might also be perceived as a bug.
Not sure what's the best TBH.

src/pages/ProfilesExplorerView/components/SceneExploreServiceLabels/components/SceneGroupByLabels/components/SceneLabelValuesGrid/domain/types.ts
grafakus: Indeed, that's a great idea but I'm not sure we should include it in this view, which is already quite complex (and might become difficult to understand for novice users).

My plan is to start prototyping a new comparison view, based on the use case that you've just shared and others.

src/pages/ProfilesExplorerView/infrastructure/helpers/getSeriesLabelFieldName.ts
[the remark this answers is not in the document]
grafakus: Here's my suggestion in the [last commit](https://github.com/grafana/explore-profiles/pull/117/commits/756d32bc27836d4d2cadb7827e6f6459551ed096):

image

PR #120, opened 2024-08-12
refactor(*): Prepare comparison flow improvements

grafakus:
### ✨ Description

**Related issue(s):** blocks to https://github.com/grafana/explore-profiles/pull/117

A preparation PR before improving the comparison flow in https://github.com/grafana/explore-profiles/pull/117.

### 📖 Summary of the changes

See the diff tab for specific comments

### 🧪 How to test?

- The build should pass


src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/SceneByVariableRepeaterGrid.tsx
```diff
@@ -49,25 +48,6 @@ const GRID_TEMPLATE_COLUMNS = 'repeat(auto-fit, minmax(400px, 1fr))';
 const GRID_TEMPLATE_ROWS = '1fr';
 const GRID_AUTO_ROWS = '240px';
 
-const DEFAULT_SORT_ITEMS_FN: SceneByVariableRepeaterGridState['sortItemsFn'] = function (a, b) {
```
grafakus, line 52: Extracted to be reused in `SceneByVariableRepeaterGrid` and `SceneLabelValuesGrid`

src/pages/ProfilesExplorerView/components/SceneLabelValuesBarGauge.tsx
```diff
@@ -53,7 +56,11 @@ export class SceneLabelValuesBarGauge extends SceneObjectBase<SceneLabelValuesBa
         return;
       }
 
-      body.setState(this.getConfig(item, state.data.series));
+      const { series } = state.data;
+
+      this.publishEvent(new EventDataReceived({ series }), true);
```
grafakus, line 61: Publishing an event that can be listened to. Helps decoupling the code responsible for removing the panels without data in `SceneByVariableRepeaterGrid`.

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/domain/sortFavGridItems.ts
```diff
@@ -0,0 +1,22 @@
...
+  if (bIsFav) {
+    return +1;
+  }
+
+  if (aIsFav) {
+    return -1;
+  }
+
+  return 0;
+};
```
simonswine, line 22: Shouldn't we also order the ones that don't have a favourite consistently?

```suggestion
  if (bIsFav && !aIsFav) {
    return +1;
  }

  if (aIsFav && !bIsFav) {
    return -1;
  }

  return a.label.localeCompare(b.label);
};
```

src/pages/ProfilesExplorerView/domain/events/EventDataReceived.ts
```diff
@@ -0,0 +1,9 @@
+import { BusEventWithPayload, DataFrame } from '@grafana/data';
+
+export interface EventDataReceivedPayload {
```
simonswine, line 3: It might be be better being explicit about what was received:

```suggestion
export interface EventSeriesReceivedPayload {
```

simonswine (approved):
LGTM.

Just having two suggestions, not fully sure if they make sense.

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/domain/sortFavGridItems.ts
grafakus: Actually no, because we want the original order to still apply.
For instance, label values are already sorted by total consumption, we're just bringing the favorites on top, we don't want to change the original order.

src/pages/ProfilesExplorerView/domain/events/EventDataReceived.ts
grafakus: Good point, I guess it's a tradeoff between a generic event and a specific event.
I'll keep it in mind for the future

PR #124, opened 2024-08-19
fix(ScenesProfileExplorer): Make labels more responsive on smaller screens

### LINKED ISSUE #116: Bug: Nit: Data source dropdown overlays Exploration
When the browser width is smaller than 1080, the exploration gets covered:

image

ifrost:
### ✨ Description

**Related issue(s):**

Fixes #116

### 📖 Summary of the changes

Just an idea how we could approach #116 by hiding the labels for the data source picker and exploration type on smaller screens. It uses `xl` theme breakpoint (~1200px) which is a bit wider than needed (~1080px) so we can adjust it.

### 🧪 How to test?

Load Explore profiles app and resize the window to below 1200px

Here's how it looks like:

https://github.com/user-attachments/assets/a51357f6-b329-4977-976e-26dd1d047eca


grafakus (approved):
Looks good, thank you!

PR #128, opened 2024-08-23
feat: create new browser history entry on some user actions 

### LINKED ISSUE #121: Using the browser back and forwards buttons don't work as expected
### Bug description

- **Grafana version:** 10.4.3
- **Plugin version:** 0.1.5

Using the browser back and forwards buttons don't work as expected: it looks like some navigation back and forth works, but not consistently.

### Steps to reproduce

For instance:

1. Go to "All services"
2. Navigate to "Profile types" then "Flame graph"
3. Click on the browser's back button

#### Actual behaviour

After clicking once on the back button, nothing happens.
After a few clicks, the browser goes to a blank tab.

#### Expected behaviour

Clicking on the back button should lead to the previous view.

ifrost:
### ✨ Description

**Related issue(s):**

Fixes #121 

### 📖 Summary of the changes

Adds basic support for browser history. A new history is created upon user actions:
- Changing exploration type
- Changing profile type
- Changing service name
- Changing group by in labels exploration

Here's how it works:

https://github.com/user-attachments/assets/9f45e476-385c-489f-b7cf-fc6b74a630f1

#### Technical details

The approach is to make sure we add new user actions to the history stack. Just before the user action handler is called we create a new empty entry in the stack and then UrlSyncManager takes care of replacing that history with any updates to variables. This will include multiple updates, and chain of updates. Any changes to URL that are not coming from the user action will be automatically added to the current history.

Alternative approaches:

1) Push history change on each variable change or every time the URL changes. 

That can work fine only when there's one variable updated upon user action. When there are multiple updates, they need to be somehow batched into single entry and it's tricky because the update is asynchronous. It's also visible during loading initial state as it usually requires multiple vars to be updated.

2) Use drilldowns

That works fine but requires some boilerplate just to support it in our case (SceneApp and SceneAppPage). It works great when it's used with native tabs coming with scenes but we don't use it for exploration type. Hence, to support drilldowns it would require even more boilerplate to create correct URLs for each exploration type (basically duplicating something that scenes handles internally with preserverUrlKeys

3) Change the variable via URL directly

Instead of calling `variable.changeValueTo` we can change the value directly by modifying the URL. Dashboards do it to handle history ([example](https://github.com/grafana/grafana/blob/771933bd7414c7c32c4ba49607f489f12398f246/public/app/features/dashboard-scene/panel-edit/PanelEditor.tsx#L73)). This way there's no need to depend on `location.partial` happening after pushing a new entry (like in this PR). The drawback might be that if a variable name is not unique it's hard to determine it's name as it's mapped by UrlSyncManager/UniqueUrlKeyMapper.

### 🧪 How to test?

It requires manual testing. Please try changing elements mentioned above (and others) and see if back/forward buttons behave as expected.


src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
```diff
@@ -154,13 +155,10 @@ export class SceneProfilesExplorer extends SceneObjectBase<SceneProfilesExplorer
   }
 
   updateFromUrl(values: SceneObjectUrlValues) {
-    const stateUpdate: Partial<SceneProfilesExplorerState> = {};
-
     if (typeof values.explorationType === 'string' && values.explorationType !== this.state.explorationType) {
-      stateUpdate.explorationType = values.explorationType as ExplorationType;
+      const type = values.explorationType as ExplorationType;
+      this.setExplorationType({ type });
```
ifrost, line 160: Switching to the getter to get the body updated correctly when going back in history

src/pages/ProfilesExplorerView/domain/variables/ProfileMetricVariable.tsx
```diff
@@ -103,6 +104,9 @@ export class ProfileMetricVariable extends QueryVariable {
   onSelect = (newValue: string) => {
     reportInteraction('g_pyroscope_app_profile_metric_selected');
 
+    if (!this.state.skipUrlSync) {
```
ifrost, line 107: Checking skipUrlSync because we use this variable with skipping in other place

src/shared/domain/url-params/useUrlSearchParams.ts
```diff
@@ -14,7 +14,7 @@ export function useUrlSearchParams() {
         newSearchParams.set(key, value);
       }
 
-      history.push({ search: newSearchParams.toString() });
+      history.replace({ search: newSearchParams.toString() });
```
ifrost, line 17: Switching to replace to avoid creating redundant history items

ifrost:
cc @torkelo @dprokop, would be great to get your feedback on this approach as well to check if we're not setting ourselves for failure and if there's a better approach.

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
```diff
@@ -225,6 +223,7 @@ export class SceneProfilesExplorer extends SceneObjectBase<SceneProfilesExplorer
     item?: GridItemData;
   }) {
     if (comesFromUserAction) {
+      prepareHistoryEntry();
```
torkelo, line 226: could this not just do a locationService.partial as well (you would get history step and update state in one action)

torkelo: this is what we do for most URL synced "view" state in core dashboards (update the URL via normal link or locatinService)

src/pages/ProfilesExplorerView/domain/variables/ProfileMetricVariable.tsx
torkelo: another variant would be

```ts
 if (this.state.skipUrlSync) {
    this.changeValueTo(newValue);
} else {
  locationService.partial({ [`var-{this.state.name}`]: newValue })
}
```

need to URL encode the value there. 

doing it this way would be compatible (not conflict) with any built-in browser history history support in UrlSyncManager / interfaces

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
ifrost: You mean to update the URL instead of changing the state and let the `updateFromUrl` handle the update? 🤔  Does it mean you need to know the final name of the variable when doing `locationService.partial()`? How do you handle it if variable name may be different after processing with `urlKeyMapper.getUniqueKey` in UrlSyncManager?

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
grafakus: 👍🏾 The only drawback is that the same body is built twice - only when landing on the page for the 1st time, if there's an `explorationType` parameter in the URL.

ifrost:
Thanks @grafakus 🙌 

> 1. In "Labels": after clicking on the "Select" panel action:
> We should also call `prepareHistory` [when changing the groupBy variable value](https://github.com/grafana/explore-profiles/blob/main/src/pages/ProfilesExplorerView/components/SceneExploreServiceLabels/components/SceneGroupByLabels/SceneGroupByLabels.tsx#L256)

Great catch! I've added prepping history directly to`selectLabel`. It's easy to make errors like this, though and forgot that variable may be change in other place, though :( I wonder if there's some other pattern that could ensure all user actions go via the same route (e.g. I noticed that select label has no user tracking, should we add it?)

Should be fixed by 584db4b03be598ca415cacded567e59f273e0069
 
> 2. In "Labels" and "Flame graph", when there are both the "Service" and the "Profile type" dropdowns.

Another nice catch! Turns out when variables are updated from the URL both `loading` and `value` property are updated straight away and options are loaded with a delay. However, once options are loaded we do not re-render the Cascader (it's forced only when `value` and `loading` prop change). It all feels a bit hacky but seems like it's all because of the way Cascader works. There's no way to control it and pass new options and value without forcing a re-render.

Changed in e55d05e6230fb9da6ed19fe8dab9fbe0b711dc5e

Alternatively we can also revert adding a new history for service and profile for now and think how to unhack Cascader first 🤔

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
torkelo: true, you would need t know the real URL key (so would only work for things where you know this). 

Was just an idea to automatically get a history step for a user action 

been making progress on https://github.com/grafana/scenes/pull/760 , can you check it out?

grafakus:
@ifrost thanks for the changes 

> I noticed that select label has no user tracking, should we add it?

The tracking [is centralized in SelectAction](https://github.com/grafana/explore-profiles/blob/main/src/pages/ProfilesExplorerView/domain/actions/SelectAction.tsx#L131)

> Alternatively we can also revert adding a new history for service and profile for now and think how to unhack Cascader first 🤔

I suppose what's lacking is to be able to control the Cascader value... In the meantime, if we end up creating a key by concatenating several values and stringifying options, I'd rather go with a simpler hack to force a render each time:

```ts
    return (
      <Cascader
        key={Math.random()}
...
```

It shouldn't be a big perf issue, wdyt?

ifrost:
> It shouldn't be a big perf issue, wdyt?

It should be fine. We may get some additional renders (2 or 3) on first paint but shouldn't have a big impact. Essentially we need to re-render Cascader when value, loading state or options change which is anyway is the only props that cause the re-render of the parent component (and hence Cascader too) so it should have almost identical outcome. Updated in c05ef5792110000c5ed3e83a816c99986ef0bcfa

grafakus (approved):

PR #119, opened 2024-08-09
feat(CompareView): Implement new Comparison view with Scenes

grafakus:
### ✨ Description

**Related issue(s):** is blocked by https://github.com/grafana/explore-profiles/pull/117

This PR introduces the new comparison view:

image

As well as the ability to select the same item for baseline and comparison in the "Labels" view:

image

### 📖 Summary of the changes

See diff tab for specific comments.

### 🧪 How to test?

- Manually, after checking this PR's branch in local

PR #129, opened 2024-08-23
fix(DiffFlameGraph): Fix the "Explain Flame Graph" (AI) feature

grafakus:
### ✨ Description

**Related issue(s):** is caused by https://github.com/grafana/explore-profiles/pull/119

This PR aims at fixing the "Explain Flame Graph" feature on the comparison view. Namely, before this PR:
- the `isDiff` prop [was not passed](https://github.com/grafana/explore-profiles/pull/119/files#diff-b648e06443c5667d8938e238e6b0a8b7579e793379e4fdd1cfc0c0191a7984e6R182) to the `AiPanel` component, leading to an incorrect AI analysis based solely on a single profile
- obtaining the queries required to fetch the profiles in DOT format relied on `parseLeftRightUrlSearchParams` which worked only for the legacy pages

### 📖 Summary of the changes

To fix the issues, this PR introduces the new `SceneAiPanel` (based on the non-Scenes `AiPanel`), used both in `SceneFlameGraph` and `SceneDiffFlameGraph`.

See the diff tab for specific comments.

### 🧪 How to test?

Manually after checking this PR branch in local:
- The analysis for a single flamegraph should work as before
- The analysis for a diff flamegraph should work

In both case, the API requests made by the browser should have the correct query(-ies) and the correct time range(s). In the case of a single flame graph, the time range must be the last one used to fetch the main timeseries. And in the case of a diff flame graph, they must be equal to the ranges selected by the user on the baseline & comparison timeseries.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneAiPanel/components/AiButton/AIButton.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneAiPanel/components/AiButton/AIButton.tsx
@@ -1,0 +1,53 @@
+import { css } from '@emotion/css';
+import { IconName } from '@grafana/data';
+import { Button, useStyles2 } from '@grafana/ui';
+import { reportInteraction } from '@shared/domain/reportInteraction';
+import React, { ReactNode } from 'react';
+
+import { useFetchLlmPluginStatus } from './infrastructure/useFetchLlmPluginStatus';
+
+type AIButtonProps = {
+  children: ReactNode;
+  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  disabled?: boolean;
+  interactionName: string;
+};
+
+export function AIButton({ children, onClick, disabled, interactionName }: AIButtonProps) {
+  const styles = useStyles2(getStyles);
+  const { isEnabled, error, isFetching } = useFetchLlmPluginStatus();
+
+  let icon: IconName = 'ai';
+  let title = '';
+
+  if (error) {
+    icon = 'shield-exclamation';
+    title = 'Grafana LLM plugin missing or not configured!';
+  } else if (isFetching) {
+    icon = 'fa fa-spinner';
+    title = 'Checking the status of the Grafana LLM plugin...';
+  }
+
+  return (
+    <Button
+      className={styles.aiButton}
+      size="md"
+      fill="text"
+      icon={icon}
+      title={isEnabled ? 'Ask FlameGrot AI' : title}
+      disabled={!isEnabled || disabled}
+      onClick={(event: React.MouseEvent<HTMLButtonElement>) => {
+        reportInteraction(interactionName);
+        onClick(event);
+      }}
+    >
+      {children}
+    </Button>
+  );
+}
+
+const getStyles = () => ({
+  aiButton: css`
+    padding: 0 4px;
+  `,
+});
--- a/src/pages/ProfilesExplorerView/components/SceneAiPanel/domain/useOpenAiChatCompletions.ts
+++ b/src/pages/ProfilesExplorerView/components/SceneAiPanel/domain/useOpenAiChatCompletions.ts
@@ -1,0 +1,153 @@
+import { llms } from '@grafana/experimental';
+import { useCallback, useEffect, useState } from 'react';
+import { Subscription } from 'rxjs';
+
+import { buildPrompts, model } from './buildLlmPrompts';
+
+// taken from "@grafana/experimental"
+type Role = 'system' | 'user' | 'assistant' | 'function';
+
+type Message = {
+  role: Role;
+  content: string;
+  name?: string;
+  function_call?: Object;
+};
+
+export type OpenAiReply = {
+  reply: {
+    text: string;
+    hasStarted: boolean;
+    hasFinished: boolean;
+    messages: Message[];
+    askFollowupQuestion: (question: string) => void;
+  };
+  retry: () => void;
+  error: Error | null;
+};
+
+export function useOpenAiChatCompletions(profileType: string, profiles: string[]): OpenAiReply {
+  const [reply, setReply] = useState('');
+  const [replyHasStarted, setReplyHasStarted] = useState(false);
+  const [replyHasFinished, setReplyHasFinished] = useState(false);
+  const [messages, setMessages] = useState<Message[]>([]);
+  const [error, setError] = useState<Error | null>(null);
+  const [subscription, setSubscription] = useState<Subscription>();
+
+  const sendMessages = useCallback((messagesToSend: Message[]) => {
+    setMessages(messagesToSend);
+
+    setError(null);
+
+    setReply('');
+    setReplyHasStarted(true);
+    setReplyHasFinished(false);
+
+    const stream = llms.openai
+      .streamChatCompletions({
+        model,
+        messages: messagesToSend,
+      })
+      .pipe(
+        // Accumulate the stream content into a stream of strings, where each
+        // element contains the accumulated message so far.
+        llms.openai.accumulateContent()
+      );
+
+    const subscription = stream.subscribe({
+      next: setReply,
+      error(e) {
+        setError(e);
+        setReplyHasStarted(false);
+        setReplyHasFinished(true);
+        setSubscription(undefined);
+      },
+      complete() {
+        setReplyHasStarted(false);
+        setReplyHasFinished(true);
+        setSubscription(undefined);
+      },
+    });
+
+    setSubscription(subscription);
+  }, []);
+
+  const askFollowupQuestion = useCallback(
+    (question: string): void => {
+      const messagesToAdd: Message[] = [
+        {
+          role: 'assistant',
+          content: reply,
+        },
+        {
+          role: 'user',
+          content: question,
+        },
+      ];
+
+      try {
+        sendMessages([...messages, ...messagesToAdd]);
+      } catch (error) {
+        setError(error as Error);
+      }
+    },
+    [messages, reply, sendMessages]
+  );
+
+  useEffect(() => {
+    if (!profiles.length || messages.length > 0) {
+      return;
+    }
+
+    const prompts = buildPrompts({
+      system: 'empty',
+      user: profiles.length === 2 ? 'diff' : 'single',
+      profileType,
+      profiles,
+    });
+
+    try {
+      sendMessages([
+        {
+          role: 'system',
+          content: prompts.system,
+        },
+        {
+          role: 'system',
+          content: prompts.user,
+        },
+      ]);
+    } catch (error) {
+      setError(error as Error);
+    }
+  }, [messages.length, profileType, profiles, profiles.length, sendMessages]);
+
+  useEffect(() => {
+    return () => {
+      if (subscription) {
+        subscription.unsubscribe();
+        setSubscription(undefined);
+      }
+    };
+  }, [subscription]);
+
+  return {
+    reply: {
+      text: reply,
+      hasStarted: replyHasStarted,
+      hasFinished: replyHasFinished,
+      messages: messages,
+      askFollowupQuestion,
+    },
+    retry() {
+      if (messages.length > 0) {
+        try {
+          sendMessages(messages);
+        } catch (error) {
+          setError(error as Error);
+        }
+      }
+    },
+    error,
+  };
+}
--- a/src/pages/ProfilesExplorerView/components/SceneAiPanel/infrastructure/ProfileApiClient.ts
+++ b/src/pages/ProfilesExplorerView/components/SceneAiPanel/infrastructure/ProfileApiClient.ts
@@ -1,0 +1,38 @@
+import { TimeRange } from '@grafana/data';
+
+import { DataSourceProxyClient } from '../../../infrastructure/series/http/DataSourceProxyClient';
+
+// dot format returns string (TODO: json format later)
+type DiffProfileResponse = string;
+
+type GetParams = {
+  query: string;
+  timeRange: TimeRange;
+  format: string; // dot (TODO: json)
+  maxNodes: number;
+};
+
+export class ProfileApiClient extends DataSourceProxyClient {
+  constructor(options: { dataSourceUid: string }) {
+    super(options);
+  }
+
+  async get(params: GetParams): Promise<DiffProfileResponse> {
+    const searchParams = new URLSearchParams({
+      query: params.query,
+      from: String(params.timeRange.from.unix() * 1000),
+      until: String(params.timeRange.to.unix() * 1000),
+      format: params.format,
+    });
+
+    if (params.maxNodes) {
+      searchParams.set('max-nodes', String(params.maxNodes));
+    }
+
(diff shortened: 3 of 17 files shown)


grafakus:
As a note for the next weeks: we need to start adding end-to-end tests to capture these kind of regressions earlier.

src/pages/ProfilesExplorerView/components/SceneAiPanel/components/AiButton/AIButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  if (error) {
+    icon = 'shield-exclamation';
+    title = 'Grafana LLM plugin missing or not configured!';
+  } else if (isFetching) {
+    icon = 'fa fa-spinner';
+    title = 'Checking the status of the Grafana LLM plugin...';
+  }
+
+  return (
+    <Button
```
grafakus, line 32: I took the opportunity to remove the "new" badge:

image

The feature has already a few months now.

src/pages/ProfilesExplorerView/components/SceneAiPanel/components/AiButton/AIButton.tsx
```diff
@@ -0,0 +1,53 @@
+import { css } from '@emotion/css';
```
grafakus, line 1: This component, as well as other files have been copied from the "shared" folder and customized for the Scenes app. Once we remove the code for the legacy Comparison pages (which is still present), we'll clean up everything.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/infrastructure/useFetchDiffProfile.ts
```diff
@@ -33,10 +33,10 @@ export function useFetchDiffProfile({
     // for UX: keep previous data while fetching -> profile does not re-render with empty panels when refreshing
     placeholderData: (previousData) => previousData,
     enabled,
-    // we use "raw" to cache relative time ranges between renders, so that only refetch() will trigger a new query
     // eslint-disable-next-line @tanstack/query/exhaustive-deps
     queryKey: [
       'diff-profile',
+      dataSourceUid,
```
grafakus, line 39: I realised I forgot to add it to the key in the previous PR 🤦🏾‍♂️

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
```diff
@@ -55,7 +60,12 @@ export class SceneFlameGraph extends SceneObjectBase<SceneFlameGraphState> {
 
       dataSubscription = newState.$data?.subscribeToState((newDataState) => {
         if (newDataState.data?.state === LoadingState.Done) {
-          timelineAndProfileApiClient.setLastTimeRange(newDataState.data.timeRange);
+          const lastTimeRange = newDataState.data.timeRange;
+
+          // For the "Function Details" feature only
+          timelineAndProfileApiClient.setLastTimeRange(lastTimeRange);
```
grafakus, line 66: We should also migrate the GitHub integration to Scenes at some point in the future.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx
```diff
@@ -141,45 +147,56 @@ export class SceneDiffFlameGraph extends SceneObjectBase<SceneDiffFlameGraphStat
...
-              enableFlameGraphDotComExport={settings?.enableFlameGraphDotComExport}
-              collapsedFlamegraphs={settings?.collapsedFlamegraphs}
+              profile={data.profile}
+              enableFlameGraphDotComExport={data.settings?.enableFlameGraphDotComExport}
+              collapsedFlamegraphs={data.settings?.collapsedFlamegraphs}
             />
           )}
         </Panel>
 
-        {sidePanel.isOpen('ai') && <AiPanel className={styles.sidePanel} onClose={sidePanel.close} />}
```
grafakus, line 182: The missing `isDiff` prop caused the feature to be broken.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx
```diff
@@ -88,6 +96,13 @@ export class SceneDiffFlameGraph extends SceneObjectBase<SceneDiffFlameGraphStat
         profile: profile as FlamebearerProfile,
         settings,
         fetchSettingsError,
+        ai: {
+          panel: aiPanel,
+          fetchParams: [
```
grafakus, line 101: We now pass the correct parameters (query and time ranges).

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
```diff
@@ -94,18 +112,20 @@ export class SceneFlameGraph extends SceneObjectBase<SceneFlameGraphState> {
...
-        title: `${serviceName} flame graph (${profileMetricType})`,
+        title: this.buildTitle(),
         isLoading: isFetchingProfileData,
         isFetchingProfileData,
         hasProfileData,
         profileData,
         settings,
+        ai: {
+          panel: aiPanel,
+          fetchParams: [{ query, timeRange: lastTimeRange }],
```
grafakus, line 127: It's now easier to understand where the last time range is used and what for (vs calling `  timelineAndProfileApiClient.setLastTimeRange(lastTimeRange)` for the GitHub integration feature).

src/pages/ProfilesExplorerView/domain/variables/FiltersVariable/FiltersVariable.tsx
```diff
@@ -47,18 +46,11 @@ export class FiltersVariable extends AdHocFiltersVariable {
 
   static Component = ({ model }: SceneComponentProps<AdHocFiltersVariable & { onChangeQuery?: any }>) => {
     const styles = useStyles2(getStyles);
+
     const { key } = model.useState();
-    const [, setQuery] = useQueryFromUrl();
 
     const query = useBuildPyroscopeQuery(model, key as string);
 
-    useEffect(() => {
```
grafakus, line 55: By removing this code and the one below in `useToggleSidePanel` we are getting rid of the `query` URL search parameters.

bryanhuhta (approved):
Tested it a variety of ways locally and everything seems to work 👍

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
```diff
@@ -43,6 +44,11 @@ export class SceneExploreDiffFlameGraph extends SceneObjectBase<SceneExploreDiff
   }
 
   onActivate() {
+    // hack to force UrlSyncManager to handle a new location
+    // this will sync the state from the URL by calling updateFromUrl() on all the time ranges (`SceneTimeRange` and our custom `SceneTimeRangeWithAnnotations`) that are defined on `SceneComparePanel`
+    // if not, landing on this view will result in empty URL search parameters (to/from and diffTo/diffFrom) which will make shareable links useless
+    locationService.partial({}, true); // replace to avoid creating history items
```
grafakus, line 50: Crazy side effect of removing https://github.com/grafana/explore-profiles/pull/129/files#diff-5db2774cbc84b771946755dd3cc399bf32509129f5bf13d2177950b07f3d6119L58, the URL search parameters for all the time ranges were wiped. 

Does anybody know a better solution (idiomatic to Scenes)?

ifrost:
Played with a bit and it looks good, I'll have a look at the code in a bit. A few comments:

---

**suggestion**: Align max-nodes used for analysis with user settings

We use hardcoded [max-nodes=100](https://github.com/grafana/explore-profiles/blob/1163777b8367dfa511758699eac7af2683da8329/src/shared/components/AiPanel/infrastructure/useFetchDotProfiles.ts#L7) for the analysis. It may yield more (or less) detailed analysis than what's visible in the graph based on user settings. Would it make sense to use user settings and cap it at 100 if uses uses more + some info about how many nodes are being processed?

---

**suggestion**: Refresh or close analysis panel when query changes

It seems like we don't close or refresh the analysis when the profile type changes (only when service changes) which may lead to stale results being shown. Would it make sense to hide the analysis or refresh it automatically when query changes (profile type or filters)?

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
ifrost: Is it needed because time range is added dynamically after the URL is synced? Maybe we could use UrlSyncManager.handleNewLocation but we need to upgrade scenes to latest version (which we should do anyway but not part of this PR)

ifrost (approved):
Nothing blocking from my end but I think we should create follow-ups on code duplication, scene updates and todos

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
grafakus: > Is it needed because time range is added dynamically after the URL is synced?

Yes

grafakus:
Thank you for the suggestions!

> **suggestion**: Align max-nodes used for analysis with user settings
> 
> We use hardcoded [max-nodes=100](https://github.com/grafana/explore-profiles/blob/1163777b8367dfa511758699eac7af2683da8329/src/shared/components/AiPanel/infrastructure/useFetchDotProfiles.ts#L7) for the analysis. It may yield more (or less) detailed analysis than what's visible in the graph based on user settings. Would it make sense to use user settings and cap it at 100 if uses uses more + some info about how many nodes are being processed?

I believe we capped it to 100:
- to limit the number of OpenAI tokens
- because it didn't provide better results with higher values

@aleks-p may is it correct?

> **suggestion**: Refresh or close analysis panel when query changes
> 
> It seems like we don't close or refresh the analysis when the profile type changes (only when service changes) which may lead to stale results being shown. Would it make sense to hide the analysis or refresh it automatically when query changes (profile type or filters)?

We do: https://github.com/grafana/explore-profiles/pull/129/files#diff-950b99b789b4fc4787eb1cf83aa00347e27cd96fbb1dcd6c9ef3bd839a45ba3bR147

src/pages/ProfilesExplorerView/components/SceneAiPanel/infrastructure/ProfileApiClient.ts
```diff
@@ -0,0 +1,38 @@
+import { TimeRange } from '@grafana/data';
+
+import { DataSourceProxyClient } from '../../../infrastructure/series/http/DataSourceProxyClient';
+
+// dot format returns string (TODO: json format later)
```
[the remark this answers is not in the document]
grafakus: It all depends.

For instance, this one has been tackled in [this PR](https://github.com/grafana/explore-profiles/pull/132/files#diff-f75b69b1b716ea9aaa14b38dadc1a2fc28fb48532caa3381adf764bacdb0736cL11).

But generally speaking, because we gave more importance to speed and flexibility, we were not strict. So far, I've used private reminders to do some code maintenance whenever I could. I now believe we should have some kind of process. Let's talk!

src/pages/ProfilesExplorerView/components/SceneAiPanel/domain/useOpenAiChatCompletions.ts
```diff
@@ -0,0 +1,153 @@
+import { llms } from '@grafana/experimental';
```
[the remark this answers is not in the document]
grafakus: [AiPanel/domain/useOpenAiChatCompletions.ts](https://github.com/grafana/explore-profiles/blob/1c2095042a4117c58f8fd9e3301797dffe006737/src/shared/components/AiPanel/domain/useOpenAiChatCompletions.ts#L2) is obsolete, so I'll remove it ASAP

src/pages/ProfilesExplorerView/components/SceneAiPanel/domain/useOpenAiChatCompletions.ts
```diff
@@ -0,0 +1,153 @@
...
+import { useCallback, useEffect, useState } from 'react';
+import { Subscription } from 'rxjs';
+
+import { buildPrompts, model } from './buildLlmPrompts';
+
+// taken from "@grafana/experimental"
```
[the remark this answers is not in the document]
grafakus: This type is not exported.

aleks-p:
> I believe we capped it to 100:
> - to limit the number of OpenAI tokens
> - because it didn't provide better results with higher values
> 
> @aleks-p may is it correct?

Yes, though I think the question is more about starting with the user's max nodes and capping that at 100. This would have an impact if the user is setting max nodes to less than 100 which is fairly unlikely but possible. In this case the AI can "talk" about nodes that the user can't see. 

The issue is a bit more subtle though. The backend determines 2 maxNodes values when making a request for a DOT profile. One is used to retrieve a flame graph and a second one to convert the flame graph to a text (DOT) report (see [this](https://github.com/grafana/pyroscope/blob/ad91b32776e236d383658271a72f44d64e4905ab/pkg/querier/http.go#L153-L162
)).

TLDR: doing something like `min(maxNodes, 100)` could make sense for cases where users configure low values (which I don't think they do). The mapping between flame graph and DOT report nodes is not 1:1 though so there could always be cases where the AI reports things that are not visible to the user.

grafakus:
> > I believe we capped it to 100:
> > 
> > * to limit the number of OpenAI tokens
> > * because it didn't provide better results with higher values
> > 
> > @aleks-p may is it correct?
> 
> Yes, though I think the question is more about starting with the user's max nodes and capping that at 100. This would have an impact if the user is setting max nodes to less than 100 which is fairly unlikely but possible. In this case the AI can "talk" about nodes that the user can't see.
> 
> The issue is a bit more subtle though. The backend determines 2 maxNodes values when making a request for a DOT profile. One is used to retrieve an intermediate flame graph and a second one to convert that flame graph to a text (DOT) report (see [this](https://github.com/grafana/pyroscope/blob/ad91b32776e236d383658271a72f44d64e4905ab/pkg/querier/http.go#L153-L162)).
> 
> TLDR: doing something like `min(maxNodes, 100)` could make sense for cases where users configure low values (which I don't think they do). The mapping between flame graph and DOT report nodes is not 1:1 though so there could always be cases where the AI reports things that are not visible to the user.

Thank you for the clarification!

PR #151, opened 2024-09-04
chore: accept pyroscope_git_session cookie name

alsoba13:
### ✨ Description

**Related issue(s):** https://github.com/grafana/pyroscope-squad/issues/171

Merge after datasources are able to forward new cookie name:
* https://github.com/grafana/grafana-com/pull/11166

We are planning to rename the `GitSession` cookie to a more convenient `pyroscope_git_session`.

### 📖 Summary of the changes

* Start handling cookies with `pyroscope_git_session`.

### 🧪 How to test?
1. Run this in local.
2. Use a local pyroscope datasource as well, and check that GitHub Integration works as is (with the `GitSession` cookie name). 
3. Change backend to send `pyroscope_git_session` instead, and check that it works as well.


src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookie.ts
```diff
@@ -1,6 +1,3 @@
-/** Name of the key for the session cookie. */
-export const GITHUB_SESSION_COOKIE_NAME = 'GitSession';
-
```
alsoba13, line 3: Maybe I'm missing some typescript features but this seemed unused to me

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
```diff
@@ -1,6 +1,8 @@
 import { GitSessionCookie } from './GitSessionCookie';
 
 const GITHUB_SESSION_COOKIE_NAME = 'GitSession';
+const NEW_GITHUB_SESSION_COOKIE_NAME = 'pyroscope_git_session';
+const ACCEPTED_GITHUB_SESSION_COOKIE_NAMES = [GITHUB_SESSION_COOKIE_NAME, NEW_GITHUB_SESSION_COOKIE_NAME];
```
bryanhuhta, line 5:
```suggestion
const GITHUB_SESSION_COOKIE_NAME = 'pyroscope_git_session';
const LEGACY_GITHUB_SESSION_COOKIE_NAME = 'GitSession';
const ACCEPTED_GITHUB_SESSION_COOKIE_NAMES = [GITHUB_SESSION_COOKIE_NAME, LEGACY_GITHUB_SESSION_COOKIE_NAME];
```

I think it makes more sense to change the old name to "legacy."

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
```diff
@@ -25,11 +27,15 @@ class InternalGitSessionCookieManager implements GitSessionCookieManager {
   }
 
   setCookie(cookie: string): void {
-    if (!cookie.startsWith(`${GITHUB_SESSION_COOKIE_NAME}=`)) {
+    if (
+      !cookie.startsWith(`${GITHUB_SESSION_COOKIE_NAME}=`) &&
+      !cookie.startsWith(`${NEW_GITHUB_SESSION_COOKIE_NAME}=`)
+    ) {
+      // If incoming cookie is not named, we use the old cookie name
```
bryanhuhta, line 34: Is this what we want? It seems to make sense to use the new cookie name if we have an unnamed cookie value.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
```diff
@@ -48,15 +54,18 @@ class InternalGitSessionCookieManager implements GitSessionCookieManager {
...
+    );
     if (cookie?.key === this.rawCookie?.key && cookie?.value === this.rawCookie?.value) {
       return;
     }
 
     cookie !== undefined ? this.setCookie(`${cookie.key}=${cookie.value}`) : this.deleteCookie();
   }
 
-  private static getCookieFromJar(jar: string, name: string): Cookie | undefined {
+  private static getCookieFromJar(jar: string, candidates: string[]): Cookie | undefined {
```
bryanhuhta, line 68: I think this method should do a full scan of the cookie jar for each candidate. It should return the cookie of the first possible match. This way we'll always use cookies of candidates first in list before candidates later in the list. For example:

```
candidates = [ 'pyroscope_git_session', 'GitSession' ]
```

First, we want to check if there's any `pyroscope_git_session` cookie in the jar. Even if we find a `GitSession` first, we should keep checking if there's a `pyroscope_git_session` later. If we don't find one, we can do another scan looking for `GitSession`.

With this current implementation, we'll accept the first of either `GitSession` or `pyroscope_git_session`, which might not be useful when we're trying to migrate to using just `pyroscope_git_session`.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/__tests__/GitSessionCookieManager.spec.ts
bryanhuhta, line 1: Perhaps we can add a test case with both `GitSession` and `pyroscope_git_session` cookies in the browser.

bryanhuhta (commented):
Looks good! I left some first pass comments, but I think this is a decent direction.

We should also make sure we're addressing what to do if we find a `GitSession` cookie already in the browser. We could:
* leave it and use it as a fallback
* rename it
* delete it

No matter what is decided, let's be clear about how we plan on handling that case 👍

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
alsoba13: I'm happy to change this to use the new name.

But note that any name we choose here is kind of wrong:
If a unnamed cookie is found, we don't know whether we have a new (pyroscope_git_session) or old (GitSession) backend.

Said that, I don't really know when we can have a unnamed cookie. Maybe I miss something.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
alsoba13: changed

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
```diff
@@ -25,50 +27,67 @@ class InternalGitSessionCookieManager implements GitSessionCookieManager {
...
   deleteCookie(): void {
-    document.cookie = `${GITHUB_SESSION_COOKIE_NAME}=; expires=Thu, 01 Jan 1970 00:00:00 UTC;`;
+    document.cookie = `${GITHUB_SESSION_COOKIE_NAME}=; Path=/; expires=Thu, 01 Jan 1970 00:00:00 UTC;`;
+    this.deleteLegacyCookie();
     this.rawCookie = undefined;
     this.sessionCookie = undefined;
   }
 
+  private deleteLegacyCookie(): void {
+    document.cookie = `${LEGACY_GITHUB_SESSION_COOKIE_NAME}=; Path=/; expires=Thu, 01 Jan 1970 00:00:00 UTC;`;
```
alsoba13, line 63: In Chrome, I had to set `Path=/;` to delete the cookie. Before using this, the cookie was still visible in the browser

alsoba13:
@bryanhuhta I did the requested changes. I'm aware this solution is not perfect yet, but still wondering if payoffs deserve the effort taking into account that I was planning to revert everything related to multi-cookie support in this PR: https://github.com/grafana/explore-profiles/pull/152

How long do you expect to have this version running before merging the cleanup? I imagine myself merging the other PR before the changes in this PR hit production

alsoba13:
We decided to change the approach and go for a full migration in one step. We'll merge both frontend and backend with the new cookie name. We don't expect to have a relevant impact on this not being fully sync, as this feature is not very popular.

Yet, we will still delete old cookie names

bryanhuhta (approved):
LGTM

PR #155, opened 2024-09-06
chore: Add and use usePrevious hook to fix side-effect errors

ifrost:
### ✨ Description

**Related issue(s):**

N/A

### 📖 Summary of the changes

Fixes a warning thrown after selecting a filter:

```
Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.
```

https://github.com/user-attachments/assets/b63a5cc9-67e4-42d7-a625-2806995bac22

We're calling blur() during the rendering which in this case should not lead to any nasty errors, but still it's a side effect so it's a bit cleaner if done after rendering finishes.

## Technical details

I've added `usePrevious` which is the same as the one coming with react-use. We can also consider adding react-use as a dependency if we take advantage of other hooks as well.

### 🧪 How to test?

* Add a quick filter 
* The error should not show up in the dev console

### DIFF
--- a/src/shared/components/QueryBuilder/QueryBuilder.tsx
+++ b/src/shared/components/QueryBuilder/QueryBuilder.tsx
@@ -1,7 +1,8 @@
 import { css, cx } from '@emotion/css';
 import { SelectableValue } from '@grafana/data';
 import { Button, useStyles2 } from '@grafana/ui';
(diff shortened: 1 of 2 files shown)


ifrost:
Blocked by https://github.com/grafana/explore-profiles/pull/157

grafakus (approved):
Thanks!

PR #171, opened 2024-09-16
fix(*): Fix useUrlSearchParams

grafakus:
### ✨ Description

**Related issue(s):** `-`

Before this PR, the URL search parameters were not correctly picked up after being updated.

We can see this issue in the "Diff flame graph" view when the default `maxNodes` is not present in the URL. In this case, its value is determined by fetching the user settings, which takes some time. And because the `useUrlSearchParams` hook was not forcing a re-render after setting the new value, the component would never display the flame graph.

### 📖 Summary of the changes

`useUrlSearchParams ` now relies on the `useLocation` hook from `react-router-dom`. As a result, the E2E tests related to the "Diff flame graph" view have now the correct screenshots (the incorrect screenshot should never have been pushed to the repo in the first place - shame on me 🤦🏾‍♂️).

### 🧪 How to test?

- The build should pass


docs/sources/choose-a-view/index.md
```diff
@@ -170,5 +170,5 @@ It normalizes the data by comparing the percentage of total time spent in each f
 This lets you compare two different queries that may have different total amounts of time spent in each function.
 
 Similar to a `git diff`, it takes the selected flame graphs and highlights the differences between them.
-<!-- from old screenshot of cpu usage profile Red represents an increase in CPU usage from the baseline to the comparison and green represents a decrease. -->
 
+<!-- from old screenshot of cpu usage profile Red represents an increase in CPU usage from the baseline to the comparison and green represents a decrease. -->
```
grafakus, line 174: Auto-lint

docs/sources/determine-use-case/index.md
```diff
@@ -59,5 +59,4 @@ After you locate the profile with a spike, select either **Profile types** view
 Illuminate performance issues across a number of dimensions when you are doing a proactive analysis.
 This could be a cost-cutting exercise or attempt to improve latency or memory usage in a non-incident scenario. This exercise could be across multiple services or within a service across profile types, labels, or a combination of the two.
 
-
```
grafakus, line 62: Auto-lint

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/CodeContainer/infrastructure/useFetchVCSFile.ts
```diff
@@ -21,7 +21,14 @@ type FetchResponse = {
   };
 };
 
-export function useFetchVCSFile({ enabled, dataSourceUid, repository, gitRef, localPath, rootPath }: FetchParams): FetchResponse {
```
grafakus, line 24: Auto-lint

src/shared/domain/url-params/useUrlSearchParams.ts
```diff
@@ -1,12 +1,11 @@
...
+import { useHistory, useLocation } from 'react-router-dom';
 
 export function useUrlSearchParams() {
   const history = useHistory();
+  const location = useLocation();
 
   return {
-    // we don't use useLocation() for performance reasons: it would take too long to update after calling history.push()
-    // and multiple history.push() would occur instead of a single one (e.g. happens when no params are present in the URL)
-    searchParams: new URLSearchParams(window.location.search),
```
grafakus, line 9: 🤦🏾‍♂️

PR #172, opened 2024-09-16
feat(AppHeader): Add Settings button

grafakus:
### ✨ Description

**Related issue(s):** `-`

| Before | After |
|  ---   |  ---  |
| image | image |

### 📖 Summary of the changes

This PR makes one more step towards a single "Explore Profiles" link in the main Grafana menu by removing the "Settings" link from the main menu in favour of a "Settings" button in the app header (grouped with the "Copy shareable URL" and "Plugin info" buttons).

Additionally,  this PR makes the "Exploration" buttons look more like a breadcrumb (see screenshot above).

### 🧪 How to test?

- The new E2E test for the "Settings" button should pass


e2e/tests/profile-types-view/profile-types.spec.ts
```diff
@@ -80,4 +85,14 @@ test.describe('Profile types view', () => {
       await expect(exploreProfilesPage.getSceneBody()).toHaveScreenshot();
     });
   });
+
+  test('Settings button', async ({ exploreProfilesPage }) => {
```
grafakus, line 89: We chose the "Profile types" view because it's not the default view, so going back with the browser button should properly set it back as the active view.

src/plugin.json
```diff
@@ -54,14 +54,6 @@
       "role": "Viewer",
       "addToNav": true,
       "defaultNav": false
-    },
-    {
-      "type": "page",
-      "name": "Settings",
-      "path": "/a/%PLUGIN_ID%/settings",
-      "role": "Admin",
```
grafakus, line 62: @bryanhuhta  Isn't it weird that we scoped this page to admins only? After all they are user settings

ifrost:
> Additionally, this PR makes the "Exploration" buttons look more like a breadcrumb (see screenshot above).

What problem does it solve? Personally I find it a bit confusing why all menu items get selected when I go straight to Diff flame graph view for example. 

Also using multiple primary buttons may be confusing to users (see https://grafana.com/developers/saga/components/buttons/button/#hierarchy--button-pairings)

matryer:
@grafakus minor feedback, I found those multiple blue call-to-action buttons a little distracting (the breadcrumbs).

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
```diff
@@ -441,12 +458,17 @@ export class SceneProfilesExplorer extends SceneObjectBase<SceneProfilesExplorer
...
               )}
 
-              <IconButton
-                name="share-alt"
-                tooltip="Copy shareable link to the clipboard"
-                onClick={actions.onClickShareLink}
-              />
-              <VersionInfoTooltip />
+              <div className={styles.miscButtons}>
+                <IconButton name="cog" tooltip="View/edit user settings" onClick={actions.onClickUserSettings} />
```
ifrost, line 462: **suggestion (non-blocking)**: Add a back/close button to Settings View

WIth this change the Settings View is going to be more integrated with the page and it feel it would be good to have close/back button on it. Once user opens it there's no clear way to go back  to main view.

grafakus: I believe they will click on the browser's back button, which will lead them where they were.

grafakus:
> > Additionally, this PR makes the "Exploration" buttons look more like a breadcrumb (see screenshot above).
> 
> What problem does it solve? Personally I find it a bit confusing why all menu items get selected when I go straight to Diff flame graph view for example.
> 
> Also using multiple primary buttons may be confusing to users (see https://grafana.com/developers/saga/components/buttons/button/#hierarchy--button-pairings)

It's to support the main [use cases](https://github.com/grafana/explore-profiles/blob/main/docs/sources/determine-use-case/index.md) and to give a sense of progression: reducing the scope from multiple services to a single flame graph

grafakus:
> @grafakus minor feedback, I found those multiple blue call-to-action buttons a little distracting (the breadcrumbs).

Are you also distracted by the arrows? I believe the arrows and the continuous selections go hand in hand.

src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
grafakus: Just in case, added in https://github.com/grafana/explore-profiles/pull/172/commits/ed1926b923c1bc9c06341afeaefc58a45c99c9fe

ifrost (dismissed):

PR #173, opened 2024-09-16
feat(*): Upgrade Grafana to v11.2.0

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR upgrades Grafana to v11.2.0

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The build should pass


package.json
```diff
@@ -35,15 +35,15 @@
...
   "dependencies": {
     "@emotion/css": "^11.1.3",
-    "@grafana/data": "10.4.3",
+    "@grafana/data": "11.2.0",
     "@grafana/experimental": "^1.7.4",
     "@grafana/faro-web-sdk": "^1.7.0",
     "@grafana/faro-web-tracing": "^1.7.0",
-    "@grafana/flamegraph": "11.1.0-179306",
-    "@grafana/runtime": "10.4.3",
+    "@grafana/flamegraph": "11.2.0",
```
grafakus, line 42: I assume that the upgrade of the flame graph package has something to do with the new height of the E2E screenshots

ifrost: A lot of screenshots show an alert banner saying "Unauthorized". Could it be a regression?

ifrost: I think it's coming from 
http://localhost:3000/api/user/preferences enpoint

ifrost: It seems to be happening because of https://github.com/grafana/grafana/pull/89908. When the menu is opened it loads preferences (which was not happening in previous versions).

ifrost (dismissed):
"Unauthorized" error is actually a bug in Grafana 11.2.0. We can merge it as it is and update when it's fixed or use 11.1.0 for time being until it's solved. I've reported the bug in here: https://github.com/grafana/grafana/issues/93388

grafakus:
> "Unauthorized" error is actually a bug in Grafana 11.2.0. We can merge it as it is and update when it's fixed or use 11.1.0 for time being until it's solved. I've reported the bug in here: [grafana/grafana#93388](https://github.com/grafana/grafana/issues/93388)

Thank you for the investigation.

package.json
grafakus: Fixed in https://github.com/grafana/explore-profiles/pull/173/commits/eae8f8fd938041261b1d2df38d9d0440cde5601c

ifrost (approved):

PR #179, opened 2024-09-18
ci: Update workflows to use actions that don't need organization secrets

jdbaldry:
Each repository can only have 100 organization secrets and there are now more than 100 in our organization which causes inconsistent behavior.

Some repositories don't have the secrets they need assigned.

These composite actions use secrets stored in Vault that are available to all repositories.

- `publish-technical-documentation-next.yml` has been tested with https://github.com/grafana/writers-toolkit/blob/main/.github/workflows/publish-technical-documentation.yml.

**Special notes for your reviewer**:

There is some copy-paste involved in the creation of these workflows. Please check:

For `publish-technical-documentation-next.yml`:

- [ ] The `on.push` `branches` and `paths` filters are correct for your repository.
- [ ] The `jobs.sync.if` repository matches your repository.
- [ ] The `jobs.sync.steps[1].with.website_directory` matches the directory you publish to in the website repository.

### DIFF
--- a/.github/workflows/publish-technical-documentation-next.yml
+++ b/.github/workflows/publish-technical-documentation-next.yml
@@ -5,33 +5,17 @@ on:
     branches:
       - main
     paths:
-      - 'docs/sources/**'
+      - "docs/sources/**"
   workflow_dispatch:
 jobs:
   sync:
     if: github.repository == 'grafana/explore-profiles'
+    permissions:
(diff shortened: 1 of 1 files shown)


knylander-grafana (approved):
The PR paths look okay to me for the docs. I'm not sure if my approval will let you merge.

PR #182, opened 2024-09-18
docs: Update README.md to call out profiles as being in preview

09jvilla:
This aligns with how we're going to message explore profiles at ObsCon, which is "in public preview for all grafana cloud and grafana open source users".


knylander-grafana (approved):
Thank you for updating the docs.

PR #189, opened 2024-09-24
refactor: Use relative URLs for API calls

### LINKED ISSUE #188: feature: Allow using the app with custom domain and subpath
### Feature description

When using following Grafana config:

```
[server]
root_url = http://localhost:3000/this-is-a-subpath/
```

And having additional mapping in `/etc/hosts`, example:

```
127.0.0.1       localhost
127.0.0.1       grafana
```

A few issues:
* Going to settings page it will open: `http://localhost:3000/this-is-a-subpath/this-is-a-subpath/a/grafana-pyroscope-app/settings` which returns 404
* When using the app from `http://grafana:3000/this-is-a-subpath` a few endpoints will attempt to load data from `localhost:3000` and failing due to cross-domain policy. To make this work `root_url` needs to be adjusted

**Describe the solution you'd like**

Ideally we could use relative paths to make sure URL are constructed correclty

### Alternatives considered

**Describe alternatives you've considered**

We can keep it as it is and require matching config though it doesn't seem to be required in other places of Grafana. 

### Additional context

Related issues:

https://github.com/grafana/explore-profiles/issues/84

ifrost:
### ✨ Description

**Related issue(s):**

Fixes #188 

### 📖 Summary of the changes

Using relative paths to Grafana API instead of pulling it directly from the config

### 🧪 How to test?

See #188 for more details.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
@@ -26,6 +26,7 @@ import { History } from 'history';
 import React, { useCallback } from 'react';
 import { useHistory } from 'react-router-dom';
 
+import { PLUGIN_BASE_URL } from '../../../../constants';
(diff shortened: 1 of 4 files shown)


grafakus (approved):
Thank you!

PR #220, opened 2024-10-10
feat(*): Add query link extension

ifrost:
### ✨ Description

**Related issue(s):**

Related to https://github.com/grafana/grafana/pull/94532

### 📖 Summary of the changes

The extension will show a link in Explore/Dashboard's QueryRow pointing to the app with data source preselected from the query row.

The link to Explore Profiles shows only when writing a Pyroscope Query. Only data source is carried over to the app link (no timerange or labels yet). 

### 🧪 How to test?

Use grafana from this branch; https://github.com/grafana/grafana/pull/94532


src/module.ts
```diff
@@ -1,6 +1,36 @@
...
+      if (!context || !context.query || !context.datasource) {
+        return undefined;
+      }
+
+      if (context.datasource.type !== 'grafana-pyroscope-datasource') {
+        return undefined;
+      }
+
+      return {
+        path: `/a/grafana-pyroscope-app/profiles-explorer?var-dataSource=${context.datasource.uid}`,
```
grafakus, line 32: Should we try to determine the service name and link to the "Profile types" view?

ifrost: Yes! Ultimately we will determine service name, profile and labels in follow-up PRs coming soon after this one. This version is bare bones to get started with all plugins 👍

src/plugin.json
```diff
@@ -1,10 +1,11 @@
 {
   "$schema": "https://raw.githubusercontent.com/grafana/grafana/main/docs/sources/developers/plugins/plugin.schema.json",
   "type": "app",
-  "name": "Explore Profiles",
+  "name": "Explore with Profiles app",
```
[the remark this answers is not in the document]
ifrost: Nope! 🙈  Updated!

grafakus (approved):

src/module.ts
ifrost: I'll update the PR to include the service name and from/to if it's provided

src/module.ts
```diff
@@ -1,6 +1,66 @@
...
+  const explorationTypeParam = `&explorationType=${explorationType}`;
+  if (timeRange) {
+    timeRangeParam = `&from=${timeRange.from}&to=${timeRange.to}`;
+  }
+
+  return `/a/grafana-pyroscope-app/profiles-explorer?${datasourceParam}${serviceNameParam}${profileTypeParam}${timeRangeParam}${explorationTypeParam}`;
+}
+
+export const plugin = new AppPlugin<AppPluginSettings>()
+  .addLink<PluginExtensionExploreContext>({
```
grafakus, line 42: Just a silly request: could we move all the code below to a different file?

By keeping the entry point `module.ts` neat, it's easier to understand and maintain (e.g. think that we will have to bring back the settings page in the future). It also pave the way for adding some focused unit tests at some point.

grafakus (dismissed):
LGTM. My only recommendation is to move the new code into a different file to keep things tidy. Besides: 👍🏾👍🏾

src/links.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const datasourceParam = `var-dataSource=${pyroscopeQuery.datasource?.uid}`;
+  const serviceNameParam = serviceName ? `&var-serviceName=${serviceName}` : '';
+  const profileTypeParam = `&var-profileMetricId=${pyroscopeQuery.profileTypeId}`;
+  const explorationTypeParam = `&explorationType=${explorationType}`;
+  if (timeRange) {
+    timeRangeParam = `&from=${timeRange.from}&to=${timeRange.to}`;
+  }
+
+  return `/a/grafana-pyroscope-app/profiles-explorer?${datasourceParam}${serviceNameParam}${profileTypeParam}${timeRangeParam}${explorationTypeParam}`;
```
grafakus, line 35: Maybe it's safer to use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams ?

grafakus, line 35: Maybe it's safer to use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams ?

grafakus (approved):

grafakus (approved):

grafakus (approved):

PR #223, opened 2024-10-15
refactor(Tracking): Use better TS types

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR just strengthen a bit our TS types related to tracking/analytics.

### 📖 Summary of the changes

See diff tab.

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/shared/domain/reportInteraction.ts
+++ b/src/shared/domain/reportInteraction.ts
@@ -1,6 +1,8 @@
 import { config, reportInteraction as grafanaReportInteraction } from '@grafana/runtime';
 
 import { PYROSCOPE_APP_ID, ROUTES } from '../../constants';
+import { LayoutType } from '../../pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/components/SceneLayoutSwitcher';
+import { PanelType } from '../../pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/components/ScenePanelTypeSwitcher';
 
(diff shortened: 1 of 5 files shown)


src/shared/components/FlameGraph/components/domain/useExportMenu.ts
```diff
@@ -10,7 +10,7 @@ import { getExportFilename } from './getExportFilename';
 /* Note: no pprof export, as the underlying API only accepts a single query (see PprofApiClient) */
 export function useExportMenu({ profile, enableFlameGraphDotComExport }: ExportDataProps) {
   const downloadPng = () => {
-    reportInteraction('g_pyroscope_export_profile', { format: 'png' });
+    reportInteraction('g_pyroscope_app_export_profile', { format: 'png' });
```
grafakus, line 13: Naming consistency. I don't believe it's a problem to change the name now.

src/shared/domain/reportInteraction.ts
```diff
@@ -21,7 +23,45 @@ function getExtraProperties() {
   return extraProperties;
 }
 
-export function reportInteraction(interactionName: string, properties?: Record<string, unknown>) {
+// hey future dev: don't forget to add any new value to our features tracking dashboard!
+export type INTERACTION_NAME =
```
ifrost, line 27: In other places we use CamelCases for type names. Any reason why this should be different?

ifrost (approved):

src/shared/domain/reportInteraction.ts
grafakus: No actually, I'll fix it. Thx!

Issue #224, opened 2024-10-16
[diff flame graph view] Clarify what the user has to do to see the diff flame graph


grafakus:
### Feature description

**Is your feature request related to a problem? Please describe.**

When landing on the "Diff flame graph view", it's not 100% clear which selections have to be made in order to view a diff flame graph, some users might end up in a situation like:

![Image](https://github.com/user-attachments/assets/9c3a59ee-4506-47ef-8f86-d631af9965b2)

Without understanding that they haven't selected the flame graph ranges on the time series:

![Image](https://github.com/user-attachments/assets/e05d9bca-205f-4910-87d1-0777274a68c4)

**Describe the solution you'd like**

1. **Improve the info message** "_Select both the baseline and the comparison flame graph time ranges to view the diff flame graph._"
Suggestion: "_In Flame graph mode, select both the baseline and the comparison flame graph time range, using your mouse, to view the diff flame graph._"
2. **Add an animated GIF** to show the user how to select the flame graph ranges
3. **Add an "Auto diff" button** which, once clicked, would select two areas with the biggest variance

### Alternatives considered

**Describe alternatives you've considered**

`-`

### Additional context

**Slack conversation:** https://raintank-corp.slack.com/archives/C03NCLB4GG7/p1729081997377229 (cc @adamquan)

Rperry2174:
I think we should do the following... we should consider two states:
https://github.com/user-attachments/assets/c5f19fcc-fdbd-4aac-8fee-be025ad3c47b

1. If the user has never selected anything and is just now landing on the diff page
2. If a user has landed on the diff page but is now selecting a **new time range**

## User has never selected anything and is just now landing on a "blank" diff page 

I believe in this case we need to teach the user that they need to select something on the timeline. We can do this via two options:
1. The user can autoselect something using our button. I think this button should pick the 5 minutes on either side of the **lowest** point on the baseline graph as the baseline.. the button should pick the 5 minutes on either side of the **highest** pint for the comparison 
2. The user can select something on the graph as instructed by the gif. See the image below for how a user should select something 

If we anted to get really fancy maybe we could even highlight the border of the line graphs panel until a user has selected something on the pink one and then highlight the border on the blue line graph panel until a user has selected both (we could also change the gif to be pink or blue depending on what they have / have not highlighted) 

## User has landed on the diff page but is now selecting a **new time range**
The whole point of us not automatically showing a flamegraph is so that we can change the time ranges independently from the flamegraph calculation. 

[...]

PR #228, opened 2024-10-18
feat(DiffFlameGraph): Add "how to" infos

grafakus:
### ✨ Description

**Related issue(s):** relates to https://github.com/grafana/explore-profiles/issues/224

This PR adds a collapsible section when landing on the "Diff flame graph" view and the user has not selected both flame graphs time ranges. The collapsed section includes an animated gif:

| Before | After |
|  ---   |  ---  |
| image | Screenshot 2024-10-18 at 11 34 11 |
|  |Screenshot 2024-10-18 at 11 34 47|

### 📖 Summary of the changes

The E2E screenshots have also been updated. See diff tab for specific comments.

### 🧪 How to test?

`-`


webpack.config.ts
```diff
@@ -46,6 +46,7 @@ const config = async (env): Promise<Configuration> => {
       extensions: ['.ts', '.tsx', '.js', '.json'],
       alias: {
         '@shared': path.resolve(__dirname, './src/shared'),
+        '@img': path.resolve(__dirname, './src/img'),
```
grafakus, line 49: Convenience

Rperry2174:
let's make the "how" uncollapsed by default -- one less click for the user

bryanhuhta:
> 1. let's make the "how" uncollapsed by default -- one less click for the user. without it being visibile by default we run the same issue of someone potentially missing it and not knowing what to do

Just my 2¢, but this is a double-edged sword. Having the section expanded by default will likely help newcomers understand how to use the diff view on their first impression. However beyond that, I expect an auto-expanded "how-to" like this would be extraordinarily annoying and disruptive beyond the first one or two page lands. Once a user understands the flow, prominently displaying a large gif on how to use the tool which you must scroll by our manually collapse will not be a great experience. This is definitely a trade-off between "good first experience" and "good general experience." Do we want to design for the new user or for the general user? I would recommend against an auto-expanded help, but I will not die on this hill.

grafakus:
I agree with Bryan, it's one more click for the user, _once_.

Let's try to make baby steps and divide & conquer. We can change and iterate later on.

PR #231, opened 2024-10-21
feat(DiffView): Add CTAs and comparison presets

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/explore-profiles/issues/224

This PR adds two call to actions when the user lands on the view:
- "Auto-select": will select half of each time ranges for the diff flame graph
- "Choose a preset": will open the presets dropdown

| Before | After |
|  ---   |  ---  |
| image | image |

Additionally, it introduces the new **Comparison presets dropdown** to help our users to quickly select different time periods:

image

Currently, we only provide 6 built-in presets:

1. Last hour (30m-window)
2. Last hour (1h-window)
3. 6h ago vs now (30m-window)
4. 24h ago vs now (30m-window)
5. Auto-select (half range)
6. Auto-select (whole range)

### 📖 Summary of the changes

See diff tab.

### 🧪 How to test?

- The build should pass
- Manually, after checking out this PR


ifrost:
> Related issue(s): relates to https://github.com/grafana/explore-profiles/issues/224

Does it only relate or actually fixes that issues? Is there anything else to do?

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/SceneComparePanel.tsx
```diff
@@ -199,6 +207,19 @@ export class SceneComparePanel extends SceneObjectBase<SceneComparePanelState> {
...
       });
     });
+
+    const timeRangeSub = this.state.$timeRange.subscribeToState((newState, prevState) => {
+      if (newState.from !== prevState.from || newState.to !== prevState.to) {
+        this.updateTitle('');
+      }
+    });
+
+    return {
```
ifrost, line 217: **suggestion (non-blocking)**: Automatically clean up subscriptions

I think in scenes when you use `this._subs.add(...)` it will automatically clean up all subscriptions when a scene is deactivated. Anything added with `this.subscribeToEvent(...)` should also be removed automatically:

- [subscribeToEvent()](https://github.com/grafana/scenes/blob/v4.22.0/packages/scenes/src/core/SceneObjectBase.tsx#L114) - [clean up code](https://github.com/grafana/scenes/blob/v4.22.0/packages/scenes/src/core/SceneObjectBase.tsx#L311)
- _subs.add [clean up code](https://github.com/grafana/scenes/blob/v4.22.0/packages/scenes/src/core/SceneObjectBase.tsx#L311)

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/SceneComparePanel.tsx
```diff
@@ -211,16 +232,65 @@ export class SceneComparePanel extends SceneObjectBase<SceneComparePanelState> {
...
+  applyPreset({ from, to, diffFrom, diffTo, label }: Preset) {
+    this.setDiffRange(diffFrom, diffTo);
+
+    this.state.$timeRange.setState(buildTimeRange(from, to));
+
+    this.updateTitle(label);
+  }
+
+  setDiffRange(diffFrom: string, diffTo: string) {
+    const $diffTimeRange = this.state.timeseriesPanel.state.body.state.$timeRange as SceneTimeRangeWithAnnotations;
```
ifrost, line 244: **thought:** Should pull the state up?

Drilling down the state looks a bit cumbersome and makes testing harder do to mocking. Do you think it would make sense / be feasible to pull `timeRange` state up and save it only in SceneComparePanel. Timeseries panel looks for nearest timeRange anyway so will use the same one. This is not needed in this PR, just food for thought.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
```diff
@@ -54,18 +59,42 @@ export class SceneExploreDiffFlameGraph extends SceneObjectBase<SceneExploreDiff
...
+    );
+
+    this._subs.add(
+      this.subscribeToEvent(EventDiffChoosePreset, () => {
+        this.state.presetsPicker.openSelect();
+      })
+    );
+  }
+
+  autoSelectDiffRanges() {
```
ifrost, line 84: **suggestion (non-blocking)**: Make the change a bit more explicit

At first I wasn't sure what and why was selected. Would it make sense to add some description on what "auto-select" actually means? (i.e. selecting first half of the first range and second half of the second range)

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx
```diff
@@ -115,7 +118,23 @@ export class SceneDiffFlameGraph extends SceneObjectBase<SceneDiffFlameGraphStat
...
+    this.publishEvent(new EventDiffAutoSelect({}), true);
+  };
+
+  onClickChoosePreset = () => {
+    reportInteraction('g_pyroscope_app_diff_choose_preset_clicked');
+
+    this.publishEvent(new EventDiffChoosePreset({}), true);
+  };
+
+  onClickLearn = () => {
```
ifrost, line 133: **todo**: It's not being used and the event is not tracked at the moment.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/ScenePresetsPicker/ScenePresetsPicker.tsx
```diff
@@ -0,0 +1,273 @@
...
+          comparison: {
+            from: 'now-1h',
+            to: 'now',
+            diffFrom: 'now-1h',
+            diffTo: 'now',
+            label: 'last hour',
+          },
+        },
+        {
+          value: '6h ago vs now',
```
ifrost, line 84: **question**: Is the rane intentional?

It selects the following (at 11:23):

Screenshot 2024-10-29 at 11 23 47

So it compares 6h-5.5h with 0.5h-0h ago

So it it was 12:00 it would compare: 6:00-6:30 with 11:30-12:00. I was kind of expecting it would compare 5:30-6:00 with 11:30-12:00 (so the "to" is shifted by 6 hours, not 5.5h. 

Same with the 24h preset below. I'm not sure what makes more sense for users, so maybe it's good the way it is. Just wanted to double check.

ifrost (changes requested):
Looks good 🙌  works as expected. One thing that is missing is one event is not being tracked. Other comments are non-blocking 👍

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
grafakus: Ok, I guess it's better to do it in `SceneComparePanel` where the logic is then

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/SceneComparePanel.tsx
grafakus: Thank you for this! I always wanted to refactor it ; I'll do it in a future PR.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/SceneComparePanel.tsx
grafakus: As we discussed face to face: this time range is actually a timerange-like object (a hack) that we use to enable annotations on the time series (see [SceneTimeRangeWithAnnotations](https://github.com/grafana/explore-profiles/blob/main/src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/components/SceneTimeRangeWithAnnotations.ts)). So it makes sense that we keep it as close as possible as where it's used (it's also responsible to change the data received from the API on-the-fly, in order to add the annotation on the time series). 

Also, if in the future the platform provides a way to add range selections out-of-the-box (via a new API or so), it's likely that it's going to be configured on the timeseries panel.

Finally, for testing, I wouldn't mock anything ; I'd use end-to-end testing as we do now because these user interactions are quite complex (because they involve many different components).

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx
grafakus: Ooops, my bad! It should be tracked.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/ScenePresetsPicker/ScenePresetsPicker.tsx
grafakus: Good question: maybe it would make more sense to center the (e.g.) 6h-mark.
I.e. selecting -6h15 to -5h45 so that the user sees 30m around the 6h mark. Having said this, I find it a bit arbitrary... I doubt that our users will select a preset to be precise about time, but more like a good approximation to start adjusting the desired range.

grafakus: I made the changes

grafakus:
@ifrost I'll merge this one and work on the follow-ups tomorrow. Thank you for the review and comments!

PR #234, opened 2024-10-22
fix(SettingsView): Fix back button after modifying the max nodes setting

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR fixes the following bug:

- Go to the "Settings" view
- Modify the maximum number of nodes
- Save the settings
- Click on the "Back to Explore Profiles" button
- Go to the "Flame graph" view

**Actual behaviour:** the new maximum number of nodes is not applied
**Expected behaviour:** the new maximum number of nodes is applied

### 📖 Summary of the changes

### 🧪 How to test?

1. The build should pass
2. Manually, after checking out this PR on your local machine.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/components/domain/useHeader.ts
+++ b/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/components/domain/useHeader.ts
@@ -3,7 +3,7 @@ import { SceneObject, SceneVariable } from '@grafana/scenes';
 import { displaySuccess } from '@shared/domain/displayStatus';
 import { useCallback } from 'react';
 import { useHistory } from 'react-router-dom';
-import { PLUGIN_BASE_URL } from 'src/constants';
+import { PLUGIN_BASE_URL, ROUTES } from 'src/constants';
 
 import { ProfilesDataSourceVariable } from '../../../../domain/variables/ProfilesDataSourceVariable';
 import { ExplorationType } from '../../SceneProfilesExplorer';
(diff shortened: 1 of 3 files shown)


src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/components/domain/useHeader.ts
```diff
@@ -62,7 +62,8 @@ export function useHeader({ explorationType, controls, body, $variables, onChang
       onClickShareLink,
       onClickUserSettings: useCallback(() => {
         reportInteraction('g_pyroscope_app_user_settings_clicked');
-        history.push(`${PLUGIN_BASE_URL}/settings`);
+
+        history.push(`${PLUGIN_BASE_URL}${ROUTES.SETTINGS}`, { referrer: window.location.href });
```
grafakus, line 66: We use the state to save the referrer.

src/pages/SettingsView/domain/useSettingsView.ts
```diff
@@ -61,7 +66,15 @@ export function useSettingsView() {
...
       },
       goBack() {
-        history.back();
+        if (!referrerRef.current) {
+          history.push(`${PLUGIN_BASE_URL}${ROUTES.PROFILES_EXPLORER_VIEW}`);
+          return;
+        }
+
+        const backUrl = new URL(referrerRef.current);
+        backUrl.searchParams.set('maxNodes', String(currentSettings.maxNodes));
```
grafakus, line 75: Because the referrer URL contains the previous max nodes value, we need to update it before going back.

PR #244, opened 2024-10-25
fix(ShareableUrl): Fix link generation when the default time range is selected

grafakus:
### ✨ Description

**Related issue(s):** `-`

When the default time range is selected and not modified, the shareable link does not work: there are no absolute `from`/`to` URL search parameters in the shareable URL. This PR fixes it.

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The unit tests should pass


marcsanmi (approved):
LGTM 👍

PR #141, opened 2024-08-28
feat(*): Add histogram visualizations

grafakus:
### ✨ Description

**Related issue(s):** solves https://github.com/grafana/explore-profiles/issues/75

image

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

Manually, after checking this PR's branch locally:
- In the "Labels" view, by clicking on the new "Histograms" button
- After favoriting a panel, it should be displayed properly in the "Favorites" view

In both "Labels" and "Favorites", the expanded panel should be displayed properly (after clicking on the "expand" panel icon):
image

PR #249, opened 2024-10-30
feat(TimeSeries): Add menu with scale options

grafakus:
### ✨ Description

**Related issue(s):** `-`

image

and after choosing the Log2 scale in the menu:

image

This feature also opens the way for the histogram visualization settings (see [previous PR comment](https://github.com/grafana/explore-profiles/pull/141#pullrequestreview-2266671181)).

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

Manually, after checking out this PR locally


src/pages/ProfilesExplorerView/components/SceneLabelValuesTimeseries.tsx
```diff
@@ -71,9 +73,55 @@ export class SceneLabelValuesTimeseries extends SceneObjectBase<SceneLabelValues
...
         .setHeaderActions(headerActions(item))
+        .setMenu(
+          new VizPanelMenu({
+            items: [
+              {
+                text: 'Scale',
+                type: 'group',
+                subMenu: [
+                  {
+                    text: 'Logarithmic (2)',
```
bryanhuhta, line 84: Does the "(2)" here refer to log base 2? If so, is that relevant? Maybe we can drop it. I spent a bit trying to figure out what "(2)" meant. If there's no other log options to select I think it makes sense to not include it.

bryanhuhta (dismissed):
LGTM 👍 

This isn't a blocker, but one observation I had when testing this is that I forgot which visualizations were in log scale and which were in linear scale. Maybe we can signal this somehow with an icon?

To emphasize this point, one of these viz is has a log scale. Can you quickly tell which one?

<details><summary>Answer</summary>
<p>
It's the third one!
</p>
</details> 

Screenshot 2024-10-30 at 3 05 55 PM

Or how about now? One of these is in log scale, do you know which one?

<details><summary>Answer</summary>
<p>
Row 2, column 2!
</p>
</details> 

Screenshot 2024-10-30 at 3 14 57 PM

src/pages/ProfilesExplorerView/components/SceneLabelValuesTimeseries.tsx
grafakus: Base 10 is also a common option, I think it's important to be specific. 
I'd go for `Logarithmic (base 2)` (vernose)  or `Log2`.

grafakus:
> LGTM 👍
> 
> This isn't a blocker, but one observation I had when testing this is that I forgot which visualizations were in log scale and which were in linear scale. Maybe we can signal this somehow with an icon?

Naive question: why is it important? Wouldn't it be more important to sync all the y-axis ranges?

PR #250, opened 2024-10-30
Use ?? operator instead of || operator

bryanhuhta:
### ✨ Description

In code view, blank lines were being rendered as `???` instead of and empty line. This PR fixes the logic to make sure blank lines are rendered correctly.

Before:

Screenshot 2024-10-30 at 2 56 12 PM

After:

Screenshot 2024-10-30 at 2 55 42 PM


grafakus (approved):
Thank you!

Issue #253, opened 2024-11-01
[feature] Optimize Auto-Select Behavior for Sub-Time-Range Comparisons in Pyroscope


Rperry2174:
### Feature description

**Is your feature request related to a problem? Please describe.**

The current "Auto-select" behavior for comparing sub-time ranges in Pyroscope can result in unintuitive or resource-heavy comparisons, especially when working with extended time ranges (e.g., 7 days). This can lead to operations that may overwhelm the database, affecting performance and usability.

**Describe the solution you'd like**

Update the "Auto-select" feature for sub-time-range comparisons to default to selecting the first and last 15% (?) of the chosen time range. This way it works when viewing a longer period (e.g., 7 days) or a shorter time period. This approach will help ensure that users can easily compare changes that occur at the beginning and end of the period without causing database performance issues.

this will ensure that IF there is a diff to be illuminated it will almost always fall in between these two time periods and the resulting flamegraph will show something interesting

**Alternatives considered**

Other potential solutions could include:
- Allowing users to customize or define their default sub-time-range for comparison.
- Providing prompts or suggestions that guide users to choose appropriate sub-time-ranges based on their specific use case.

PR #254, opened 2024-11-04
fix(DiffView): Clicking on "Auto-select" selects a 25% range

grafakus:
### ✨ Description

**Related issue(s):** fixes https://github.com/grafana/explore-profiles/issues/253

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

Manually:
- clicking on "Auto-select" should always select a 25% range on both the baseline & the comparison time series
- when the 25% are bigger than one day, the range should be capped to 1 day


colega:
Should we cap auto-select to an absolute value instead? Like `1d` max. 
If I'm looking at 14d range, selecting a 25% of that still would be a lot.

ifrost (approved):
LGTM 👍  but as mentioned above it'd be better to cap at some absolute value (e.g. 1 day)

grafakus:
> Should we cap auto-select to an absolute value instead? Like `1d` max. If I'm looking at 14d range, selecting a 25% of that still would be a lot.

Done!

ifrost (approved):

Issue #283, opened 2024-11-18
[bug] Service name from the URL is being overridden by last used service name


ifrost:
### Bug description

- **Grafana version:** 11.3.0
- **Plugin version:** 0.1.16

### Steps to reproduce

1. Go to "Profile Types" view
2. Ensure you have at least two services available
3. Select a service name from the dropdown
4. Change the value of `YOUR_SERVICE_NAME` in the URL `http://.../a/grafana-pyroscope-app/profiles-explorer?...var-serviceName=YOUR_SERVICE_NAME...` to a different service name
5. Load the page with the new URL

https://github.com/user-attachments/assets/2b98da2f-cad1-4c21-b6d5-f39f6dd7d788

#### Actual behaviour

The previously used service name is used

#### Expected behaviour

The service name from the URL should be used

PR #284, opened 2024-11-18
fix(ProfileTypesView): Load last used service name only if it's not provided in the URL

ifrost:
### ✨ Description

**Related issue(s):**

Fixes: #283

### 📖 Summary of the changes

In https://github.com/grafana/explore-profiles/pull/222 we introduced loading last used service name from local storage. The stored value is applied during `onActivate()` (it was changed in https://github.com/grafana/explore-profiles/pull/233). This however overrides the service name value if it's provided in the URL. The change is to skip loading last used service name if it's already set (e.g. from the URL)

### 🧪 How to test?

See reproduction steps in #283


grafakus (approved):

grafakus:
LGTM, thanks for the fix!

Just a detail: the fix is for profile types, labels and flame graph views.

PR #287, opened 2024-11-19
feat(*): Upgrade Grafana to v11.3.0

ifrost:
### ✨ Description

**Related issue(s):** `-`

This PR upgrades Grafana to v11.3.0

### 📖 Summary of the changes

All references upgraded from 11.2.0 to 11.3.0

### 🧪 How to test?

- The build should pass


src/shared/ui/PluginInfo.tsx
```diff
@@ -14,18 +14,24 @@ const { buildInfo: grafanaBuildInfo } = config;
...
-    </div>
-  );
+  if (context) {
+    const {
+      meta: {
+        info: { version, updated },
+      },
+    } = context;
+
+    return (
```
grafakus, line 26: Good catch.
Instead of removing _entirely_ the header, why not gracefully degrade?
At least by displaying a question mark or "information not available", so it wouldn't be perceived as a bug from the user perspective.

src/app/components/Routes/Routes.tsx
```diff
@@ -13,13 +13,13 @@ export function Routes() {
 
   return (
     <Switch>
-      <Route path={prefixRouteWithPluginBaseUrl(ROUTES.PROFILES_EXPLORER_VIEW)} exact>
```
grafakus, line 16: Why the change?

ifrost: `exact` param is not available in react router 6:

```
ERROR in ./src/app/components/Routes/Routes.tsx:22:67
TS2322: Type '{ children: Element; path: string; exact: true; }' is not assignable to type 'IntrinsicAttributes & RouteProps'.
  Property 'exact' does not exist on type '(IntrinsicAttributes & PathRouteProps) | (IntrinsicAttributes & LayoutRouteProps)'.
    20 |         <AdHocView />
    21 |       </Route>
  > 22 |       <Route path={prefixRouteWithPluginBaseUrl(ROUTES.SETTINGS)} exact>
       |                                                                   ^^^^^
    23 |         <SettingsView />
    24 |       </Route>
    25 |       {/* Default Route */}
```

from [docs](https://reactrouter.com/en/main/upgrading/v5#relative-routes-and-links):

> `<Route exact>` is gone. Instead, routes with descendant routes (defined in other components) use a trailing * in their path to indicate they match deeply

src/shared/ui/PluginInfo.tsx
ifrost: Good idea! How about [this](https://github.com/grafana/explore-profiles/pull/287/commits/0b3f2e1b337e42155c2c650e95383d575b32edb6)?

Screenshot 2024-11-19 at 14 36 24

src/app/components/Routes/Routes.tsx
grafakus: Then we have to prevent discrepancies between the local environment/devx and the other environments by also upgrading React Router in the npm dependencies.

ifrost: I have reverted the change and added explicit react-router version in package.json. Grafana 11.3.0 provides v5. Locally if we don't specify react-router it starts pulling v6 to node_modules 🤔

grafakus: > it starts pulling v6 to node_modules
Transitive dependencies FTW!

grafakus (approved):
👍🏾👍🏾

Issue #242, opened 2024-10-25
[diff flame graph view] Add a time range sync button


grafakus:
### Feature description

**Is your feature request related to a problem? Please describe.**

It's currently difficult to quickly select the same ranges on both the comparison and baseline time series

**Describe the solution you'd like**

Add a "Sync time ranges" toggle, like we have in class Explore:

![Image](https://github.com/user-attachments/assets/9578aa60-351d-44a1-abee-955c90676b67)

Once activated, whether the user is in the "Time picker" or "Flame graph" selection mode, both selections are in sync.

### Alternatives considered

**Describe alternatives you've considered**

`-`

### Additional context

`-`

PR #288, opened 2024-11-20
feat(DiffView): Time ranges sync

grafakus:
### ✨ Description

**Related issue(s):** solves https://github.com/grafana/explore-profiles/issues/242

This PR adds two "Sync time ranges" buttons in the diff view:

image

They allow the selections on the baseline and the comparison time series to stay in sync. Once entered "sync mode", data refreshes are also synced.

### 📖 Summary of the changes

**The general idea is:**
- The new sync buttons emit an event that is handled by the `SceneExploreDiffFlameGraph` parent object, which activates them in the UI and force a first sync (the one clicked is used as reference)
- When synced, the `SceneComparePanel` objects will emit events whenever the time range or the diff time range change (it is also done via events)
- `SceneExploreDiffFlameGraph` is responsible for synchronizing both time ranges when receiving these events

The PR also includes some refactoring (moving the TS `CompareTarget` enum where it makes more sense).

### 🧪 How to test?

- The new E2E tests should pass


ifrost:
Looks great! A few things were a bit confusing to me:

* I set the range on the left chart, then I clicked "sync" button. I kind of expected the same range to be applied to the other chart but nothing happend. Ideally when clicking on the "sync" button I would like the range of the current chart to be applied to the other one
* Why presets are not available when sync is enabled? Presets are out-of sync but selecting a preset could simply remove sync option
* It would be nice to have sync option in the URL. I was checking the graph, refreshed the page to get fresh results (I know I can click on run query button but ctrl+R is quicker) and the sync was gone. I am also biased because in Classic Explore sync option is saved in the URL.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
```diff
@@ -79,6 +81,51 @@ export class SceneExploreDiffFlameGraph extends SceneObjectBase<SceneExploreDiff
...
+
+        comparisonPanel.toggleTimeRangeSync(enable);
+        baselinePanel.toggleTimeRangeSync(enable);
+
+        this.state.presetsPicker.toggle(enable);
+      })
+    );
+
+    this._subs.add(
+      // eslint-disable-next-line sonarjs/cognitive-complexity
```
ifrost, line 98: Should we tweak the rule instead of skipping it?

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
```diff
@@ -79,6 +81,51 @@ export class SceneExploreDiffFlameGraph extends SceneObjectBase<SceneExploreDiff
         this.state.presetsPicker.openSelect();
       })
     );
+
+    this._subs.add(
+      this.subscribeToEvent(EventEnableSyncTimeRanges, (event) => {
+        const { enable } = event.payload;
+        const { baselinePanel, comparisonPanel } = this.state;
+
+        comparisonPanel.toggleTimeRangeSync(enable);
```
ifrost, line 90: The state is duplicated - it's always true/false in both. Would it make sense to keep only one variable to store it? Or maybe sync it with URL with a single variable?

grafakus:
> Looks great! A few things were a bit confusing to me:
> 
> * I set the range on the left chart, then I clicked "sync" button. I kind of expected the same range to be applied to the other chart but nothing happend. Ideally when clicking on the "sync" button I would like the range of the current chart to be applied to the other one
> * Why presets are not available when sync is enabled? Presets are out-of sync but selecting a preset could simply remove sync option
> * It would be nice to have sync option in the URL. I was checking the graph, refreshed the page to get fresh results (I know I can click on run query button but ctrl+R is quicker) and the sync was gone. I am also biased because in Classic Explore sync option is saved in the URL.

Great suggestions!
- "_When clicking on the "sync" button I would like the range of the current chart to be applied to the other one_": will do!
- "_Presets are out-of sync but selecting a preset could simply remove sync option_": I agree
- "_It would be nice to have sync option in the URL_": here there are 2 things: 
  1. If you're in sync mode: why not syncing both refresh as well? Refreshing one would refresh the other
  2. Storing the option in the URL: IMO it's not a critical parameter required to display a correct UI when landing on the page. But maybe one way to think about it (remembering that the URL parameters are a public interface/API that we have to maintain) is to ask ourselves: "if we store it, deploy to prod and then, after some time, we remove it, would our users complain?" I don't think so :)

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
grafakus: 🤔 I'm not sure to follow... Unless we move it to the URL (but I don't think we should - see my main comment), the state has to be duplicated: both panels need to know if their sync button is active or not.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/SceneExploreDiffFlameGraph.tsx
grafakus: I realize I can refactor the code to prevent the complexity.

src/shared/ui/PluginInfo.tsx
```diff
@@ -22,7 +23,10 @@ function InfoMenuHeader() {
 
   return (
     <div className={styles.menuHeader}>
-      <h5>🔥 Explore profiles v{version}</h5>
```
grafakus, line 25: Not related to this PR

ifrost (approved):
Great stuff!

PR #289, opened 2024-11-20
chore(Versioning): Bump the minor version by default

aleks-p:
We currently bump the patch version for every release. This is counterintuitive because significant additions and changes are treated the same as smaller bugfixes. A user thinking of upgrading the plugin might think a change is small, when in reality it might be bigger.

From semver.org:

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backward compatible manner
> PATCH version when you make backward compatible bug fixes

As far as I understand, right now we manually select what portion of the version to bump when creating a new tag. This PR changes the default selection in that flow to "minor". Alternatively we could leave the default empty and force a selection, however I think that "minor" should be the usual selection.


grafakus:
HI @aleks-p ! Thanks for pointing this out. The reasoning was that, because Explore Profiles is in public preview, the versioning will "keep on patching"... Once we go GA (at end of the Q), we'll release v1.0 and adopt semantic versioning.

aleks-p:
thanks @grafakus

As discussed separately, this can wait. However, there is also no harm (and there are benefits) to do it for pre-v1 versions since we already expose the version numbers publicly.

ifrost:
> This is counterintuitive because significant additions and changes are treated the same as smaller bugfixes. A user thinking of upgrading the plugin might think a change is small, when in reality it might be bigger.

I think it makes sense to use minor as default. However, if we want to follow semver, we need to be mindful about the version we pick so I would also be happy with not having any default and always enforce think-first. Unless we automate versioning based on conventional commits.

Note: All our current flows are continuous so we get, e.g. 0.1.0 -> 0.2.0 -> 0.2.1 -> 0.3.0. At this point we don't have automation for making 0.2.2 based of 0.2.1 (though maybe we won't need it).

grafakus:
@aleks-p we got the chance to talk a bit longer about versioning with @ifrost today.

I'll dump my notes in a Google doc but the PR looks good as it is so I'll merge, thank you!

PR #290, opened 2024-11-20
refactor(Tracking): Better error logging for Scenes variables

grafakus:
### ✨ Description

**Related issue(s):** `-`

When a Scene variable encounters an error, the Scenes library [only sets the error message](https://github.com/grafana/scenes/pull/371/files#diff-4e4e615e280869407aa645bac1e449fe8a23bbb28381092800f00d3dc8a4bd7cR234) in the variable state. This PR ensures we log an error with Faro, and not only the message.

Additionally, there's a of behaviour for the "Labels" view: we prevent the labels data source to fail when fetching label values fail.

### 📖 Summary of the changes

The logger calls are moved to the data sources instead of the React components. See diff tab for specific comments.

### 🧪 How to test?

`-`


src/pages/ProfilesExplorerView/infrastructure/labels/LabelsDataSource.ts
```diff
@@ -73,6 +73,49 @@ export class LabelsDataSource extends RuntimeDataSource {
...
+  async fetchLabelValues(query: string, from: number, to: number, labelName: string, variableName?: string) {
+    let values;
+
+    try {
+      values = await labelsRepository.listLabelValues({ query, from, to, label: labelName });
+    } catch (error) {
+      logger.error(error as Error, {
+        info: 'Error while loading Pyroscope label values!',
+        variableName: variableName || '',
+      });
```
grafakus, line 100: Here we don't rethrow, we degrade gracefully so the UI can render something (hopefully) useful for the user.

PR #304, opened 2024-11-29
refactor(*): Use the Intl API for sorting items

grafakus:
### ✨ Description

**Related issue(s):** see https://github.com/grafana/grafana/pull/96862 for performance impact

A PR just to modernize our code base by using the [Intl APIs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator).

### 📖 Summary of the changes

See diff tab.

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/domain/sortFavGridItems.ts
+++ b/src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/domain/sortFavGridItems.ts
@@ -1,3 +1,5 @@
+import { localeCompare } from '@shared/domain/localeCompare';
+
 import { FavAction } from '../../..//domain/actions/FavAction';
 import { FavoritesDataSource } from '../../../infrastructure/favorites/FavoritesDataSource';
 import { GridItemData } from '../types/GridItemData';
@@ -7,7 +9,7 @@ export const sortFavGridItems: (a: GridItemData, b: GridItemData) => number = fu
   const bIsFav = FavoritesDataSource.exists(FavAction.buildFavorite(b));
 
   if (aIsFav && bIsFav) {
-    return a.label.localeCompare(b.label);
+    return localeCompare(a.label, b.label);
   }
 
   if (bIsFav) {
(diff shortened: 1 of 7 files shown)


src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/domain/sortFavGridItems.ts
```diff
@@ -7,7 +9,7 @@ export const sortFavGridItems: (a: GridItemData, b: GridItemData) => number = fu
   const bIsFav = FavoritesDataSource.exists(FavAction.buildFavorite(b));
 
   if (aIsFav && bIsFav) {
-    return a.label.localeCompare(b.label);
+    return localeCompare(a.label, b.label);
```
ifrost, line 12: Wouldn't this be exactly the same? Accroding to [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare):

> The localeCompare() method of [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) values returns a number indicating whether this string comes before, or after, or is the same as the given string in sort order. In implementations with [Intl.Collator API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) support, this method simply calls Intl.Collator.

ifrost: Oh sorry, just read the paragraph below which says:

> When comparing large numbers of strings, such as in sorting large arrays, it is better to create an [Intl.Collator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) object and use the function provided by its [compare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/compare) method.

:D

grafakus: No worries :)

Issue #314, opened 2024-12-11
Empty datasource query fails


kolesnikovae:
### Bug description

- **Grafana version:** Grafana Enterprise v11.5.0-80050 (production)
- **Plugin version**: v0.1.17 Commit SHA: 31687553

Exploring an empty datasource causes query failures.

### Steps to reproduce

1. Add an empty Pyroscope datasource to Grafana.
2. Open Explore Profiles.
3. Close the "Welcome to Explore Profiles" window.
4. Observe query failure.

![Image](https://github.com/user-attachments/assets/cff50c54-d062-42f3-b2bf-7e5240a85dba)

#### Expected behaviour

1. No queries are sent if the datasource reports no data for the tenant.
2. No query errors are observed.

#### Context

This is Pyroscope v2 backend, which is supposed to be fully backward-compatible at the API level.

1. In the dev console I see that `GetProfileStats` returns an empty response (as expected):

![Image](https://github.com/user-attachments/assets/fe660f89-e9fb-43c1-ab17-0691817b0312)

2. After that, the datasource issues `Series` request that yields no data (as expected):

![Image](https://github.com/user-attachments/assets/a6151ff8-32b7-46c1-892b-39000cd08004)

3. Finally, Pyroscope's backend is requested with malformed queries that cannot succeed, and the error message is shown to the end user as is:

![Image](https://github.com/user-attachments/assets/3ab81bf2-1606-46b4-8809-df955b2d213e)

1. I'm wondering if this is expected, given that both `GetProfileStats` and `Series` report no data.
2. The queries in p.3 are invalid:
  * The `{service_name=""}` label selector cannot result a non-empty response.
  * The `profileTypeId` cannot be an empty string.

I'm wondering if this is intended behaviour, and we should implement workarounds on the Pyroscope side (simply ignore queries with empty `profileTypeId`), or if this is something we want to fix in the Explore Profiles plugin.

grafakus:
Hi @kolesnikovae, thank you for reporting and for all the details.

Here's a PR to prevent it: https://github.com/grafana/explore-profiles/pull/316

kolesnikovae:
Awesome, thanks @grafakus!

PR #316, opened 2024-12-11
fix(QueryRunners): Prevent invalid queries to run

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/explore-profiles/issues/314

| Before | After |
|  ---   |  ---  |
| image | image|

### 📖 Summary of the changes

This PR adds an activation handler to each query runner being instantiated. The handler is validating the query being passed so that, if either no service name or profile type is defined, no query is executed.

### 🧪 How to test?

See the the linked issue above:
- The Network tab should not show any `query=ds=xxx` (failing) HTTP requests.
- The UI should not show any error icon on the timeseries panels, "No data" should be displayed instead

Be careful to remove the last selected service name from the browser's local storage before testing. Make also sure that the URL parameters have been removed before loading the page.

### DIFF
--- a/src/pages/ProfilesExplorerView/infrastructure/withPreventInvalidQuery.ts
+++ b/src/pages/ProfilesExplorerView/infrastructure/withPreventInvalidQuery.ts
@@ -1,8 +1,28 @@
 import { LoadingState } from '@grafana/data';
 import { sceneGraph, SceneQueryRunner } from '@grafana/scenes';
+import { parseQuery } from '@shared/domain/url-params/parseQuery';
+import { logger } from '@shared/infrastructure/tracking/logger';
 
 export function withPreventInvalidQuery(queryRunner: SceneQueryRunner) {
   queryRunner.addActivationHandler(() => {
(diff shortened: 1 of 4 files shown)


src/pages/ProfilesExplorerView/infrastructure/withPreventInvalidQuery.ts
```diff
@@ -0,0 +1,58 @@
...
+import { logger } from '@shared/infrastructure/tracking/logger';
+
+export function withPreventInvalidQuery(queryRunner: SceneQueryRunner) {
+  queryRunner.addActivationHandler(() => {
+    const { profileTypeId, labelSelector } = queryRunner.state.queries[0];
+
+    if (!profileTypeId) {
+      queryRunner.setState({
+        queries: [{ refId: 'null' }],
+        data: buildErrorData(queryRunner, 'Missing profile type!'),
```
grafakus, line 13: `[{ refId: 'null' }]` prevents the query to be executed but, unfortunately, even if we build some synthetic "errored" data, the UI doesn't display the error message.

Issue #115, opened 2024-08-06
bug: We display unit with prefix


simonswine:
![image](https://github.com/user-attachments/assets/5e252ef8-3fc0-4db7-af26-3a323bfa8626)

![image](https://github.com/user-attachments/assets/c9e03d0d-d08c-4341-9b63-a855dd3d1e4b)

Reported by @bboreham 

* We should not use the SI prefix, as the axis scales it to the fitting SI prefix and that's confusing
* Total might be not accurate, as it shows it over time

Unsure if this is the best alternative:

* "Seconds of CPU time consumed"
* "Bytes allocated in the heap"

grafakus:
In some cases, we could even get rid of the unit completely, e.g. "CPU time consumed".

We will review [all the descriptions](https://github.com/grafana/explore-profiles/blob/main/src/shared/infrastructure/profile-metrics/profile-metrics.json) and make a suggestion.

Related: https://github.com/grafana/pyroscope/issues/2740

PR #318, opened 2024-12-12
chore: Update profile type descriptions

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/explore-profiles/issues/115

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

`-`


src/shared/infrastructure/profile-metrics/profile-metrics.json
```diff
@@ -43,28 +43,28 @@
...
-    "description": "Total objects allocated",
+    "description": "Number of objects allocated",
     "type": "alloc_objects",
     "group": "memory",
     "unit": "short"
   },
   "memory:alloc_space:bytes:space:bytes": {
     "id": "memory:alloc_space:bytes:space:bytes",
-    "description": "Total bytes allocated in the heap",
+    "description": "Size of memory allocated in the heap",
```
knylander-grafana, line 53: Do we need to state that the number is in bytes?

src/shared/infrastructure/profile-metrics/profile-metrics.json
```diff
@@ -43,28 +43,28 @@
...
     "type": "alloc_space",
     "group": "memory",
     "unit": "bytes"
   },
   "memory:inuse_objects:count:space:bytes": {
     "id": "memory:inuse_objects:count:space:bytes",
-    "description": "Count of objects currently in use",
+    "description": "Number of objects currently in use",
```
knylander-grafana, line 60: Do number and count imply different actions or are they good synonyms? I know in some context a count implies a different action than number. That might not matter here.

src/shared/infrastructure/profile-metrics/profile-metrics.json
```diff
@@ -78,28 +78,28 @@
...
-    "description": "Total bytes allocated during CPU time",
+    "description": "Size of memory allocated during CPU time",
     "type": "alloc_size",
     "group": "alloc_size",
     "unit": "bytes"
   },
   "process_cpu:cpu:nanoseconds:cpu:nanoseconds": {
     "id": "process_cpu:cpu:nanoseconds:cpu:nanoseconds",
-    "description": "Total nanoseconds of CPU time consumed",
+    "description": "Time of CPU consumed",
```
knylander-grafana, line 102: Do we need to state that the time is in nanoseconds? also, does the updated wording have the same meaning as the original? What about this?
```suggestion
    "description": "CPU time consumed (in nanoseconds)",
```

src/shared/infrastructure/profile-metrics/profile-metrics.json
grafakus: It seems to be the confusing part (mentioned in the linked GitHub issue):
> We should not use the SI prefix, as the axis scales it to the fitting SI prefix and that's confusing

src/shared/infrastructure/profile-metrics/profile-metrics.json
```diff
@@ -113,28 +113,28 @@
...
-    "description": "Count of lock acquisitions attempted during CPU time",
+    "description": "Number of lock acquisitions attempted during CPU time",
     "type": "lock_count",
     "group": "locks",
     "unit": "short"
   },
   "process_cpu:lock_time:nanoseconds:cpu:nanoseconds": {
     "id": "process_cpu:lock_time:nanoseconds:cpu:nanoseconds",
-    "description": "Cumulative time spent acquiring locks, in nanoseconds",
+    "description": "Cumulative time spent acquiring locks",
```
knylander-grafana, line 123: Again, do we want to specify nanoseconds to help people understand what they are looking at? Or do you think it's immediately obvious?

src/shared/infrastructure/profile-metrics/profile-metrics.json
```diff
@@ -113,28 +113,28 @@
...
-    "description": "Total number of process samples collected",
+    "description": "Number of process samples collected",
     "type": "samples",
     "group": "process_cpu",
     "unit": "short"
   },
   "process_cpu:samples:count:cpu:nanoseconds": {
     "id": "process_cpu:samples:count:cpu:nanoseconds",
-    "description": "Count of samples collected over CPU time",
+    "description": "Number of samples collected over CPU time",
```
knylander-grafana, line 137: Same question about nanoseconds here. It would be something to add, even though it wasn't present before.

knylander-grafana (dismissed):
Thank you for updating the descriptions. Do we want to be consistent about mentioning how a value is presented? For example, specifying whether something is in bytes or nanoseconds.

src/shared/infrastructure/profile-metrics/profile-metrics.json
knylander-grafana: Okay. We don't have to specify bytes. What does "SI prefix" refer to?

grafakus: Oh I think I misunderstood: https://www.nist.gov/pml/owm/metric-si-prefixes

We might want to specify the unit - but ensure we don't mention the prefix ("kilo", "mega", ..)

src/shared/infrastructure/profile-metrics/profile-metrics.json
grafakus: Good question, I don't know 🤷🏾‍♂️

src/shared/infrastructure/profile-metrics/profile-metrics.json
grafakus: In this case, it's a short number (see below in the JSON: `"unit": "short"`),

src/shared/infrastructure/profile-metrics/profile-metrics.json
simonswine: Means the same to me

src/shared/infrastructure/profile-metrics/profile-metrics.json
```diff
@@ -1,70 +1,70 @@
...
-    "description": "Current number of goroutines",
+    "description": "Number of goroutines",
     "type": "goroutine",
     "group": "goroutine",
     "unit": "short"
   },
   "memory:alloc_in_new_tlab_bytes:bytes::": {
     "id": "memory:alloc_in_new_tlab_bytes:bytes::",
-    "description": "Total bytes allocated inside Thread-Local Allocation Buffers (TLAB)",
+    "description": "Size of memory allocated inside Thread-Local Allocation Buffers (bytes)",
```
simonswine, line 32: I personally find the TLAB bit useful. As that's how I know it and not "Thread-Local Allocation Buffers".

src/shared/infrastructure/profile-metrics/profile-metrics.json
simonswine: I find specifying units is tricky, and I would consistently remove them (bytes, nanoseconds).

 This is why:

The headline says it is nanoseconds, but the graph's axis is actually say the correct seconds:

image

e2e/fixtures/pages/ExploreProfilesPage.ts
```diff
@@ -361,4 +361,10 @@ export class ExploreProfilesPage extends PyroscopePage {
     await panel.hover({ position: coordsTo });
     await this.mouse.up();
   }
+
+  clickOnSyncTimerangesButton(target: 'baseline' | 'comparison') {
```
grafakus, line 365: DRY

e2e/fixtures/pages/ExploreProfilesPage.ts
```diff
@@ -342,8 +342,8 @@ export class ExploreProfilesPage extends PyroscopePage {
 
     const overlay = this.getByTestId('data-testid TimePicker Overlay Content');
 
-    await overlay.getByTestId('data-testid Time Range from field').fill(from);
     await overlay.getByTestId('data-testid Time Range to field').fill(to);
+    await overlay.getByTestId('data-testid Time Range from field').fill(from);
```
grafakus, line 346: Slight detail to prevent temporary validation error ;)

Issue #237, opened 2024-10-25
[diff flame graph view] Preserve the context when leaving and entering the view again


grafakus:
### Feature description

**Is your feature request related to a problem? Please describe.**

When the user selects some filters and changes the time ranges then leaves the view and comes back to it, their selections are lost.

**Describe the solution you'd like**

The app should preserve the selections that were made

### Alternatives considered

**Describe alternatives you've considered**

`-`

### Additional context

**Currently, the logic is:**

- when the user goes to the flame graph view, adds some filters and changes the time range then goes to the diff view, the app preserves the filters and time ranges from the flame graph view. The same thing happens from the labels view to the diff view
- when the users leaves the diff view, all the time ranges are cleared: the time pickers are set to the main time picker value (the one present in the app header in the other views) and the flame graph ranges are cleared

grafakus:
cc @adamquan

PR #319, opened 2024-12-12
fix(DiffFlameGraphView): Preserve context after leaving/re-entering the view

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/explore-profiles/issues/237

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

Manually:
1. Verify that the filters, the time pickers and the diff ranges are preserved after leaving and re-entering the view
2. Go to the "Labels" view, select a label, select 2 items for comparison and then click on the "Compare" button: the current time range and filters should be used to initialize the diff view


bryanhuhta (approved):
Okay this is really cool. I really like the how if I apply filters on the "Labels" view, then I compare, I not only get a diff view of the two labels I selected for comparison, but ALSO the original "Labels" filter applies. Nice 💯

PR #322, opened 2024-12-18
refactor(*): Lazy load page components

grafakus:
### ✨ Description

**Related issue(s):** `-`

### DIFF
--- a/src/app/components/Routes/Routes.tsx
+++ b/src/app/components/Routes/Routes.tsx
@@ -2,10 +2,11 @@ import React from 'react';
 import { Routes as ReactRouterRoutes, Route } from 'react-router-dom';
 
 import { ROUTES } from '../../../constants';
-import { AdHocView } from '../../../pages/AdHocView/AdHocView';
-import { ProfilesExplorerView } from '../../../pages/ProfilesExplorerView/ProfilesExplorerView';
-import { SettingsView } from '../../../pages/SettingsView/SettingsView';
 import { useNavigationLinksUpdate } from './domain/useNavigationLinksUpdate';
+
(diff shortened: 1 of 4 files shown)


grafakus:
Thanks for taking the time to look at the PR!

> I had to approve the drone build - not 100% sure why.

We don't have to, it's a relic of the past. We only use the Drone "tag" build whenever we want to manually test the PR.

> Interesting that you have a settings view but not a settings pages. I suppose the reason here is it's easier to get to the view rather than find the settings page via the plugins page.

That's one reason indeed (it was more discoverable in the menu than in the general plugins page), but it's not the main one:
- We wanted the ability to store user settings, like max flame graph nodes, etc., in a persistent way, because whenever a instance restarts, the plugin settings were also reset 


Note that the Settings view uses the Settings API, which is very flexible (we can store any key-values) and it's scoped to the tenant.

joey-grafana:
Cool, thanks for the added context.

| We don't have to, it's a relic of the past. We only use the Drone "tag" build whenever we want to manually test the PR.

ahh I thought it was running the tests no? That are required to pass for the PR to then be merged.

| We wanted the ability to store user settings (like max flame graph nodes, etc.) in a persistent way, because whenever a instance restarts, the plugin settings were also reset.

Wow, if the plugins settings are lost every time the instance restarts, what's the point in having the settings page (for any other plugin)!

grafakus:
> ahh I thought it was running the tests no? That are required to pass for the PR to then be merged.

In a PR, the [CI GitHub workflow](https://github.com/grafana/explore-profiles/blob/main/.github/workflows/ci.yml) runs them (Frontend job)

> Wow, if the plugins settings are lost every time the instance restarts, what's the point in having the settings page (for any other plugin)!

Yes, at least it was the case when we had the plugin Settings page in the past. It may have changed since.

PR #323, opened 2024-12-18
revert: "perf(*): Lazy load page components"

grafakus:
Reverts grafana/explore-profiles#322 after deploying in ops and seeing:

image

Will need to check our build pipeline if we want this optimization.


ifrost:
Do you have any more details what exactly was the error? File could not be loaded or it was a runtime error? 🤔  I pinned [the version with lazy loading](https://github.com/grafana/explore-profiles/commit/202d40e9b3a2d28e164180e34ad91330e38de9ad) to my dev instance and it loaded fine.

Screenshot 2025-01-03 at 10 59 58

Issue #35, opened 2024-04-26
Make the scope of the GitHub integration login clear


simonswine:
The way the Grafana data source password are scoped to the whole Grafana instance, it is easy to think that authorizing GitHub might also allow other users to access your GitHub repo's source code.

We need to make it clearer, that this affects this Browser session only, next to "Connect to"

![image](https://github.com/grafana/pyroscope-app-plugin/assets/223048/d5c42190-ab14-4ff7-afbe-82cc2704487b)

PR #328, opened 2024-12-26
feat(GitHubIntegration): Add info tooltip on connect button

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/explore-profiles/issues/35

image

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

`-`


ifrost (dismissed):

ifrost:
@simonswine do you think it's good enough to solve #35?

simonswine:
I think it seems good enough, but I am not too sure if it is correct as it is currently worded:

My understanding is that when you login for: grafana/backend-enterprise, it will also work in the same session for grafana/secret-project, or any other github project.

grafakus:
> I think it seems good enough, but I am not too sure if it is correct as it is currently worded:
> 
> My understanding is that when you login for: grafana/backend-enterprise, it will also work in the same session for grafana/secret-project, or any other github project.

New message: "_Once connected, the GitHub code will be accessible only from this browser session._"

simonswine (approved):
LGTM

ifrost (approved):

PR #347, opened 2025-01-23
fix(GitHubIntegration): Handle function details for inlining

simonswine:
When this code looked up function detail, it only took a look at the first line of a location. A location with multiple lines happens when functions are inlined, the last entry represents the caller into which the preceding entries were inlined into.

This corrects this problem and allows to find function detail in this case as well.

Reported by @colega


src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/domain/__tests__/convertPprofToFunctionDetails.spec.ts
```diff
@@ -384,7 +420,7 @@ describe('convertPprofToFunctionDetails(fnName, profile)', () => {
               [134, { cum: 1650000000, flat: 1650000000, line: 134 }],
               [120, { cum: 560000000, flat: 0, line: 120 }],
               [137, { cum: 820000000, flat: 820000000, line: 137 }],
-              [142, { cum: 400000000, flat: 10000000, line: 142 }],
```
simonswine, line 387: This is expected, as the sample contains inlining hence the spec was slightly incorrect.

simonswine:
> Refactor this function to reduce its Cognitive Complexity from 11 to the 8 allowed

I think refactoring this to be simpler will make it harder to understand IMHO, but I can give it a try. I guess I can for a start group all those arguments into a type

bryanhuhta (approved):
LGTM 👍

Issue #351, opened 2025-01-24
[Labels view] Sort label values by their max value


colega:
### Feature description

**Is your feature request related to a problem? Please describe.**

I'm exploring profiles of different pods trying to find examples of goroutine spikes. 

There's no easy way to find them, I just learned that I can click this button _outside_ of the label view:

![Image](https://github.com/user-attachments/assets/220d6181-cd61-4f0b-b9c8-fea757fb8c67)

But what I'd really want is to sort the labels by their max value in the presented range.

**Describe the solution you'd like**

In the label view, i.e. here:

<img width="1381" alt="Image" src="https://github.com/user-attachments/assets/273532a2-c179-415d-b6c8-203fe84b0f39" />

I would like to be able to sort the values by their max value.

### Alternatives considered

**Describe alternatives you've considered**

Let me run PromQL on that data.

PR #341, opened 2025-01-22
feat(LabelsView): update main time series when a "group by" label is selected

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/explore-profiles/issues/197, is blocked by https://github.com/grafana/explore-profiles/pull/219

This PR makes the main time series in the Labels view responsive to "group by" changes, for example:

image

And after interacting with the "Include"/"Exclude" buttons (including 3 out of the 52 nodes available):

image

### 📖 Summary of the changes

General overview:

- The drawer component has been lifted up to `SceneExploreServiceLabels`
- `SceneMainServiceTimeseries` reacts to the `groupBy` variable changes by updating the item stored in  `SceneLabelValuesTimeseries`'s state
- `SceneLabelValuesTimeseries` run new queries whenever the item's `queryRunnerParams.groupBy` value changes
- The height of the timeseries in both the Labels view and the Flame graph view has been increased from 200px to 240px (this required to update the E2E screenshots)
- New E2E tests have been added to verify that the main timeseries is reset properly when selecing a different service/profile type

See diff tab for specific comments.

### 🧪 How to test?

- The build should pass
- ⚠️ Manually, by ensuring that the use cases described in https://github.com/grafana/explore-profiles/issues/197 are working (see the "Additional context" section)

PR #361, opened 2025-01-29
feat(LabelsView): Add maxima visualizations

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/explore-profiles/issues/351, is blocked by https://github.com/grafana/explore-profiles/pull/341

image

This PR adds a new visualization in the "Labels" view for maxima, which can be useful to discover peaks of memory usage for instance. This should work well combined with https://github.com/grafana/explore-profiles/pull/341 and the "Include"/"Exclude" panel actions

As a small addition, the maximum value is also displayed in every timeseries panel, when a single series is rendered.

### 📖 Summary of the changes

See diff tab. I took the opportunity to regenerate all the E2E screenshots.

### 🧪 How to test?

- The build should pass


ifrost:
Just to double check. The issue says:

> In the label view, i.e. here:
> 
> I would like to be able to sort the values by their max value.

But rather than sorting values in that view by maximum we provide a different view to see max values in a table (without a way to select the label and investigate further). Is it good enough to solve the problem?

src/pages/ProfilesExplorerView/components/SceneLabelValuesTable.tsx
```diff
@@ -0,0 +1,167 @@
...
+                options: {
+                  reducers: ['max'],
+                  labelsToFields: true,
+                },
+              },
+              {
+                id: DataTransformerID.filterFieldsByName,
+                options: {
+                  exclude: {
+                    names: ['Field'],
```
ifrost, line 54: Why do we need to exclude a field called `Field`? 🤔

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/infrastructure/data-transformations.ts
```diff
@@ -2,6 +2,8 @@ import { DataFrame } from '@grafana/data';
 import { merge } from 'lodash';
 import { map, Observable } from 'rxjs';
 
+export type StatName = 'totalSeriesCount' | 'allValuesSum' | 'maxValue';
```
ifrost, line 5: unused?

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/infrastructure/data-transformations.ts
```diff
@@ -13,10 +15,19 @@ export const addStats = () => (source: Observable<DataFrame[]>) =>
     map((data: DataFrame[]) => {
       const totalSeriesCount = data?.length;
 
+      // TODO: in case of a groupBy query, find a way to always add a rank to each label value (based on allValuesSum) so that we can use it as startColorIndex to
```
ifrost, line 18: Do we have an issue for this TODO? When should we tackle it?

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/infrastructure/data-transformations.ts
```diff
@@ -13,10 +15,19 @@ export const addStats = () => (source: Observable<DataFrame[]>) =>
...
+      // always display each series consistently in the same color regardless of it's timseries, bar gauges with sums, or tables with maxima
       return data?.map((d) => {
+        let maxValue = Number.NEGATIVE_INFINITY;
+
         const allValuesSum = d.fields
           ?.find((field) => field.type === 'number')
-          ?.values.reduce((acc: number, value: number) => acc + value, 0);
+          ?.values.reduce((acc: number, value: number) => {
```
ifrost, line 25: The function is growing a bit, should be add a unit test for it?

grafakus:
> Just to double check. The issue says:

> > I would like to be able to sort the values by their max value.
 
> But rather than sorting values in that view by maximum we provide a different view to see max values in a table (without a way to select the label and investigate further). Is it good enough to solve the problem?

Yes, I pinged some colleagues and confirm it's a good solution to the problem. I agree that we should consider adding some sorting, but:
1. It will take longer to develop
2. It will require some research ; I'm not sure the solution proposed in the issue is the right thing to do. Maybe instead of sorting, the app should allow the user to directly select an element in the "Maxima" table to see a flame graph? Or even to select two elements directly in the table to see a diff flame graph? If it's the right thing to do, the grid displayed after selecting a breakdown should have less features, so adding sorting wouldn't be needed.

src/pages/ProfilesExplorerView/components/SceneLabelValuesTable.tsx
grafakus: The result of the"reduce" transformation yields:

image

where "cpu" is repeated on each row

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/infrastructure/data-transformations.ts
grafakus: good catch!

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/infrastructure/data-transformations.ts
grafakus: What do you think about reviewing them before GA?
I think reviewing them (as well as the `eslint-disable`, console warnings, unit test outputs, deprecations, ...) once per quarter is a good frequency. We could create an issue to be sure we tackle them in the future.

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/infrastructure/data-transformations.ts
grafakus: They are covered by the E2E tests I guess (this PR also updates the screenshots).

src/pages/ProfilesExplorerView/components/SceneByVariableRepeaterGrid/infrastructure/data-transformations.ts
ifrost: Sure, we can review them after GA 👍  Personally if I add a TODO to code I try to create a corresponding GitHub issue and link in the code so it's easy to plan, figure out the context and timeline.

ifrost:
> It will require some research ; I

Could we also add some feature tracking to see how this feature is being used?

grafakus:
> > It will require some research ; I
> 
> Could we also add some feature tracking to see how this feature is being used?

It is in place, the interaction name is `g_pyroscope_app_panel_type_changed`

ifrost (approved):
Thanks for all explanations 👍  LGTM!

PR #363, opened 2025-01-29
feat(Settings): Add collector rules management

simonswine:
This PR adds collector rules to the Explore profiles app.

The Pyroscope settings API stores those rules and will generate Alloy configuration (PR: https://github.com/grafana/pyroscope/pull/3865/), that then can be used by an extension of the Collector app (PR: 
https://github.com/grafana/grafana-collector-app/pull/639, thanks for your help @jarben).

Also see the design document: [DesignDoc: Allow to remote control of profiling targets from Grafana Cloud](https://docs.google.com/document/d/1LCY5jiqoRdapJ8B91QLLCbCEpnpsc9-sTAFjnR3Fe54/edit?tab=t.0)

Note: This also tries to use generated APIs from the backend (as per #354), it still uses JSON as the transport protocol though. Happy to remove this if this is not wanted. This increases the Assert sizes:

```
Before: modules by path ../node_modules/ 1.84 MiB 184 modules
Before: modules by path ./ 913 KiB (javascript) 602 KiB (asset)
After:  modules by path ../node_modules/ 2.24 MiB 221 modules
After:  modules by path ./ 929 KiB (javascript) 602 KiB (asset)
```

https://github.com/user-attachments/assets/af7da2b6-7be9-476c-bcea-f5f6729dabfd


simonswine:
There has been a lot general discussion about if this work is going forward or not. I have in the meantime reworked the UX as per our disucssion with @grafakus.

I don't plan to work further until we know more, I would still like to get a bit of feedback on the approach if you got some time.

src/pages/SettingsView/components/CollectorSettings/components/ViewRule/ViewRule.tsx
```diff
@@ -0,0 +1,117 @@
...
+              />
+            )}
+          </Stack>
+        </Collapse>
+      </Card.Description>
+      <Card.Meta>
+        {[`${data.rule.services.length} services`, `Last updated: ${formatDate(data.rule.lastUpdated)}`]}
+      </Card.Meta>
+      <Card.SecondaryActions>
+        {/* TODO(simonswine): Unsure if this is the best way of doing this */}
```
simonswine, line 81: Is there are more controlled way of linking to other parts of Grafana?

src/pages/SettingsView/components/CollectorSettings/domain/useCollectorSettings.ts
```diff
@@ -0,0 +1,91 @@
...
+    saveRule(rule: UpsertCollectionRuleRequest): Promise<void>;
+    deleteRule(ruleName: string): Promise<void>;
+  }
+> {
+  const { rules, isFetching, error: fetchError, upsertAsync, deleteAsync } = useFetchCollectorRules();
+
+  const [currentState, setCurrentState] = useState<GetCollectionRuleResponse[]>(rules ?? []);
+  const existingRuleNames = useMemo(() => currentState.map((r) => r.name), [currentState]);
+
+  // TODO: useMemo, because it's not a side effect
```
simonswine, line 29: I don't understand that, I want/need the rules to be set from the inital request, but then want to update them as we save/delete rules. How do I achieve this with `useMemo`?

src/shared/types/DomainHookReturnValue.ts
```diff
@@ -2,3 +2,9 @@ export type DomainHookReturnValue<T = any, V = any> = {
   data: Record<string, T>;
   actions: Record<string, V>;
 };
+
+// TODO: Christian is struggling with type safefty of my domain hook return values, is this a good idea?
```
simonswine, line 6: I had basically no safety with `DomainHookReturnValue`. Unsure if I am just not aware how to make the returns typed with it. But this saved me a lot of trouble

simonswine:
This has been done as extension for Grafana Cloud only

PR #364, opened 2025-01-29
ci: Test development flow changes

ifrost:
### ✨ Description

**Related issue(s):**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
     # Plugin is already signed in frontend job so if you need to use to locally you can just download the artifact
     # When pushed to main it uses "gcs-no-approval" environment which can be triggered only from main
     # to push the package automatically without approval
-    name: Package and publish plugin
+    name: Upload to GCS
     needs: [end-to-end]
     environment: ${{ github.event_name == 'push' && 'gcs-no-approval' || 'gcs' }}
     runs-on: ubuntu-latest
@@ -186,7 +186,7 @@ jobs:
           workload_identity_provider: ${{ secrets.WORKLOAD_IDENTITY_PROVIDER }}
           service_account: ${{ secrets.GCS_SERVICE_ACCOUNT }}
 
-      - name: Publish to GCS
+      - name: Upload to GCS
         uses: 'google-github-actions/upload-cloud-storage@v1'
         with:
           path: ./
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-existing-version.yml
```diff
@@ -0,0 +1,33 @@
+name: Deploy existing version
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version prefixed with v, e.g. v1.0.8 or a commit sha'
```
grafakus, line 7:
```suggestion
        description: 'GIT ref (commit SHA or tag, e.g. v1.0.8)'
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+      environment:
+        required: true
+        type: string
+        description: dev, ops or prod
```
grafakus, line 12: Could it support multiple values? I'm thinking having to rollback all environments at once, or ops and dev.

.github/workflows/ci.yml
```diff
@@ -143,7 +143,7 @@ jobs:
     # Plugin is already signed in frontend job so if you need to use to locally you can just download the artifact
     # When pushed to main it uses "gcs-no-approval" environment which can be triggered only from main
     # to push the package automatically without approval
-    name: Package and publish plugin
+    name: Upload to GCS
```
grafakus, line 146: NIT: let's use human friendly names (i.e. "Frontend build" and "E2E tests" for instance):

image

grafakus: Also, in the confirmation modal, we should display "Approve and upload". I saw "Approve and deploy"... too late! And my heart went ❣️

.github/workflows/deploy.yml
grafakus: It looks like it is already, right?

image

ifrost: Not possible natively but we can use some tricks with booleans though the code will look a bit clunky. But should be possible.

ifrost: btw. there's a feature request for it https://github.com/actions/runner/issues/2076 so please upvote

ifrost: hmm... right we can also do it by guarding it with environment 🤔

.github/workflows/ci.yml
ifrost: "Approve and deploy" is unfortunately native GitHub actions modal that shows to which environment you want to deploy. Github environment != Grafana environment. I have created "GCS" Github environment so we don't upload every commit and we can manually decide which one is uploaded.

Alternatively we can have a separate workflow that is triggered on Pull Request comment. For example you comment "/upload" in the comment of the PR and that triggers a workflow to upload to GCS

ifrost:
>. github/workflows/plugin-release.yml

Yes this one is not used. I was planning to clean up everything at the end but this one can be removed: https://github.com/grafana/explore-profiles/pull/372

> .github/workflows/weekly-release.yml

This is still part of weekly releases flow, though it's not being used. But this one would require cleaning up Argo workflow as well (which triggers this action based on cron config) so we need to tackle is separately.

grafakus (dismissed):
LGTM, nice work! 🙌🏾

grafakus (approved):

PR #367, opened 2025-01-29
ci: Prevent signing and publishing the plugin in forked repos

### LINKED ISSUE #366: ci: Forks are unable to run CI pipeline
When a PR is created from a fork then signing a plugin doesn't work as it requires to be run from a Grafana owned repo. The signing is needed only for artifacts that will be published to Plugins Catalog or Cloud so we can remove this step from forks (e.g. adding conditional step `if: ${{ !github.event.pull_request.head.repo.fork }}`) or by running it only in main (e.g. `if: github.event_name == 'push'`)

For more details check https://raintank-corp.slack.com/archives/C0734VDPSPM/p1738170674457229

Signing the plugin was added in https://github.com/grafana/explore-profiles/pull/308

ifrost:
### ✨ Description

Fixes #366

### 📖 Summary of the changes

Make sure signing the plugin happens only in main to allow running CI in forks.

### 🧪 How to test?

Run the job, e.g. https://github.com/grafana/explore-profiles/actions/runs/13041161764/job/36383129066 to see signing is being skipped:

Screenshot 2025-01-29 at 22 57 28

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,9 @@ jobs:
         run: npx @grafana/levitate@latest is-compatible --path src/module.ts --target @grafana/data,@grafana/ui,@grafana/runtime
 
       # The plugin is signed here so it's possible to use the artifact produced by the job directly
+      # However we do it only for build running from main to allow run it in forks (see https://github.com/grafana/explore-profiles/issues/366)
       - name: Setup plugin signing
+        if: github.event_name == 'push'
         uses: grafana/shared-workflows/actions/get-vault-secrets@main
         with:
           vault_instance: ops
@@ -73,6 +75,7 @@ jobs:
 
       # create MANIFEST in dist
       - name: Sign plugin
+        if: github.event_name == 'push'
         run: yarn sign
         env:
           GRAFANA_ACCESS_POLICY_TOKEN: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN }}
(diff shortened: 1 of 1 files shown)


simonswine (approved):
LGTM

Just to double check, we still want the unsigned plugin to be uploaded?

https://github.com/grafana/explore-profiles/blob/bfe49f860657cc931cbe4f4fb5172de34aae7502/.github/workflows/ci.yml#L83-L88

ifrost:
> Just to double check, we still want the unsigned plugin to be uploaded

Yep, we want to upload them for testing pull requests in dev, for such cases I think we don't need the plugin to be signed.

grafakus:
> > Just to double check, we still want the unsigned plugin to be uploaded
> 
> Yep, we want to upload them for testing pull requests in dev, for such cases I think we don't need the plugin to be signed.

Just to triple-check: are we sure uploading an unsigned plugin to a dev instance and restarting it works?

ifrost:
> > > Just to double check, we still want the unsigned plugin to be uploaded
> > 
> > 
> > Yep, we want to upload them for testing pull requests in dev, for such cases I think we don't need the plugin to be signed.
> 
> Just to triple-check: are we sure uploading an unsigned plugin to a dev instance and restarting it works?

Oh boy, It thought it would but it doesn't seem to work! I'll change it so these steps are skipped for forks and we don't even upload it to GCS for forks

grafakus (approved):
👍🏾

PR #374, opened 2025-02-03
fix(Filters): Ensure filters are always cleared when changing the data source or the service

grafakus:
### ✨ Description

**Related issue(s):** `-`

Before this PR, the filters were not always reset when changing the data source.

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

Manually, for instance:

1. Go to the "Diff flame graph" view and add baseline and comparison filters
2. Go to the "All services" view
3. Change the data source
4. Go back to the "Diff flame graph" view
5. The filters have been cleared

Also:
- The filters should be cleared when changing the service in the "Labels" view and coming back to the "Diff flame graph" view

### DIFF
--- a/src/pages/ProfilesExplorerView/domain/variables/ProfilesDataSourceVariable.ts
+++ b/src/pages/ProfilesExplorerView/domain/variables/ProfilesDataSourceVariable.ts
@@ -2,6 +2,8 @@ import { DataSourceVariable } from '@grafana/scenes';
 import { ApiClient } from '@shared/infrastructure/http/ApiClient';
 import { userStorage } from '@shared/infrastructure/userStorage';
 
+import { FiltersVariable } from './FiltersVariable/FiltersVariable';
+
 export class ProfilesDataSourceVariable extends DataSourceVariable {
   constructor() {
     super({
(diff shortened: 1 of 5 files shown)


src/pages/ProfilesExplorerView/domain/variables/ProfilesDataSourceVariable.ts
```diff
@@ -25,6 +27,8 @@ export class ProfilesDataSourceVariable extends DataSourceVariable {
         const storage = userStorage.get(userStorage.KEYS.PROFILES_EXPLORER) || {};
         storage.dataSource = newState.value;
         userStorage.set(userStorage.KEYS.PROFILES_EXPLORER, storage);
+
+        FiltersVariable.resetAll(this);
```
ifrost, line 31: I feel like resetting variables should be part of `SceneProfilesExplorer`. It already does some resetting when exploration type changes, it also contains all variables so could listen to changes in `ServiceNameVariable` and `ProfileMetricVariable` and reset `FiltersVariables`. It would be more explicit and we could avoid coupling `ProfilesDataSourceVariable`/`ServiceNameVariable` with `FiltersVariable`. What do you reckon?

grafakus: Good point, I've the changes

PR #376, opened 2025-02-03
ci: Refactor actions to make it easier to reuse them

ifrost:
### ✨ Description

**Related issue(s):** #346 

### 📖 Summary of the changes

- Moves build and package jobs to separate re-usable workflows. They will be used in manual release workflow.
- Adds creating md5 checksum to uploaded builds. This will be needed by action publishing to the catalog. ([here where it's gonna be used](https://github.com/grafana/plugin-ci-workflows/blob/24d0e3c9bf72ea25ffc12da78fe5f72075150e42/actions/plugins/publish/publish/publish.sh#L107))
- Uploads a packaged artifacts so they can be later attached to a GitHub release ([like this](https://github.com/grafana/explore-profiles/blob/ifrost/manual-release-workflow/.github/workflows/create-new-version.yml#L94C14-L94C37))

### 🧪 How to test?

- Current workflows should work the same way (e.g. https://github.com/grafana/explore-profiles/actions/runs/13135390904)
- There should be md5 file uploaded to GCS ([example](https://storage.googleapis.com/grafana-pyroscope-app/releases/grafana-pyroscope-app-06afa45808f4740a56e90897f0077f4b1473e993.zip.md5))
- There should be a new artifact available from the packaging stage (see build-package at the bottom of https://github.com/grafana/explore-profiles/actions/runs/13135390904)

Example flow: https://github.com/grafana/explore-profiles/actions/runs/13135390904


grafakus (approved):
Nice!

PR #384, opened 2025-02-06
fix(Settings): Prevent warning to be displayed when no settings is returned by the Settings API

grafakus:
### ✨ Description

**Related issue(s):** `-`

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/shared/infrastructure/settings/settingsApiClient.ts
+++ b/src/shared/infrastructure/settings/settingsApiClient.ts
@@ -13,14 +13,10 @@ class SettingsApiClient extends ApiClient {
       .fetch('/settings.v1.SettingsService/Get', { method: 'POST', body: JSON.stringify({}) })
       .then((response) => response.json())
       .then((json: ApiResponse) => {
-        if (!json.settings) {
-          return {};
-        }
-
-        const setting = json.settings.find(({ name }) => name === SettingsApiClient.PLUGIN_SETTING_NAME);
+        const setting = json.settings?.find(({ name }) => name === SettingsApiClient.PLUGIN_SETTING_NAME);
 
         if (!setting) {
-          throw new Error('Setting not found!');
+          return {};
         }
 
         return JSON.parse(setting.value);


src/shared/infrastructure/settings/settingsApiClient.ts
```diff
@@ -13,14 +13,10 @@ class SettingsApiClient extends ApiClient {
...
-        if (!json.settings) {
-          return {};
-        }
-
-        const setting = json.settings.find(({ name }) => name === SettingsApiClient.PLUGIN_SETTING_NAME);
+        const setting = json.settings?.find(({ name }) => name === SettingsApiClient.PLUGIN_SETTING_NAME);
 
         if (!setting) {
-          throw new Error('Setting not found!');
+          return {};
```
ifrost, line 19: In what case it could happen? Could we log it to faro?

grafakus: For the same reasons as above, in the code that was removed: the API could return `settings` with an empty array or - as seen this morning, settings which contains only a list of exported metrics without any plugin settings. These are edge cases that could occur when visiting the plugin for the 1st time so we prevent the user to see a warning toast when landing.

In both case, IMO the UI should be flexible _in this aspect_. Now, if the API request fails or if storing new settings fail, the UI should display an error message and it should be logged to Faro. Wdyt?

ifrost: > the API could return settings with an empty array or - as seen this morning, settings which contains only a list of exported metrics without any plugin settings

If this is an actual API error why don't we let the customer know that settings are not loaded?

> These are edge cases that could occur when visiting the plugin for the 1st time so we prevent the user to see a warning toast when landing

Or is it more of an expected / by-design behavior that when loading the plugin for the first time settings are not returned until they are saved? Could we populate empty settings in the backend upon first retrieval?

grafakus: > If this is an actual API error why don't we let the customer know that settings are not loaded?

It's not an error when the user lands for the first time on the app, they shouldn't see a warning toast. By design the Settings API is flexible. I understand your concern, we can strengthen the contract later IMO.

ifrost (approved):

Issue #295, opened 2024-11-26
Migrate Manual Release CI Workflow to GitHub Actions


ifrost:
Manual Release Workflow changes:

[Migration Plan](https://docs.google.com/document/d/1PaSzBLmDgYiWBLjurODupn334uRmmPVQc72LG6iMHYY/edit?tab=t.0#heading=h.rhfrcg6mc2ap)

Make sure version.ts and package.json are updated accordingly

Triggering the flow remains the same. 

Required changes:
 * [x] Create dev/ops/prod environments - all need review to trigger actions
 * [x] Reuse build and package from https://github.com/grafana/explore-profiles/issues/294
 * [x] GitHub: Publish latest.zip to GCS (skipped)
 * [x] GitHub: Publish {TAG}.zip to GCS
 * [ ] GitHub: Create GitHub release
 * [x] GitHub: Trigger deploy-plugin-dev argo (dev environment)
 * [x] GitHub: Trigger deploy-plugin-ops argo (ops environment)
 * [ ] GitHub: Trigger deploy-plugin-prod argo (prod environment)
 * [ ] Drone: Stop build packages on version tags (vX.Y.Z)
 * [ ] Drone: Stop publishing to GSC on version tags
 * [ ] Drone: Stop creating GitHub Releases
 * [ ] Drone: Stop publishing to Grafana.com
 * [ ] Drone: Stop deployment steps triggered by promotion

PR #387, opened 2025-02-06
ci: Disable e2e when creating a new version

ifrost:
### ✨ Description

**Related issue(s):** #295 

### 📖 Summary of the changes

Newly created flow to create a new version runs redundant e2e tests that should be passing in main.

### 🧪 How to test?

E2E tests should be run for this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,9 @@ on:
         required: false
         type: string
         description: commit sha or tag that will be checked out (use it if an existing tag needs to be checked-out)
+      e2e:
+        required: true
+        type: boolean
 
 jobs:
   frontend:
@@ -89,6 +92,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     needs: [frontend]
+    if: ${{ inputs.e2e == 'true' }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -12,6 +12,9 @@ on:
         required: false
         type: string
         description: commit sha or tag that will be checked out (use it if an existing tag needs to be checked-out)
+      e2e:
```
grafakus, line 15: I have the feeling it shouldn't be an option, it feels like we're giving too many possibilities....
Why not just disabling E2E when creating a new version because:
- the E2E were launched when merging to main
- we can assume it was deployed to dev and tested manually
- it was deployed to ops and tested by internal users
Wdyt?

ifrost: > I have the feeling it shouldn't be an option...

It's not an option for the user (you cannot decide if you want to run it or not). It's a config for reusable option because e2e is not part of that reusable workflow.

Would you prefer to split that workflow into 2 separate reusable workflows (1 - build frontend, 2 - run e2e)? Or just move e2e to ci as part of the job (so e2e won't be reusable)

grafakus: My bad, I misread the code 🤦🏾‍♂️

It's good as it is then!

grafakus (approved):

PR #389, opened 2025-02-06
ci: Release existing version

ifrost:
### ✨ Description

**Related issue(s):** #295

### 📖 Summary of the changes

This extracts a part of create-new-version.yml flow to allow carrying on with https://github.com/grafana/explore-profiles/actions/runs/13179001686/job/36785488283. Might be worth to keep it in case create-new-version.yml fails at any stage.


.github/workflows/release-existing-tag.yml
```diff
@@ -0,0 +1,87 @@
+name: Release exiting tag
```
grafakus, line 1:
```suggestion
name: Release existing tag
```

.github/workflows/release-existing-tag.yml
```diff
@@ -0,0 +1,87 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version (with v prefix, e.g. v0.1.20)'
```
grafakus, line 7:
```suggestion
        description: 'Git tag identifying the version including the "v" prefix (e.g. v0.1.20)'
```

.github/workflows/release-existing-tag.yml
```diff
@@ -0,0 +1,87 @@
+name: Release existing tag
```
grafakus, line 1:
```suggestion
name: Release existing version
```

grafakus: If you agree with this suggestion, we have to rename the file

grafakus (dismissed):

grafakus (approved):

PR #360, opened 2025-01-29
feat: Profiles to metrics UI

bryanhuhta:
### ✨ Description

**Related issue(s):**

https://github.com/grafana/pyroscope-squad/issues/340

This adds new UI elements to support creating metrics from profiles.

> [!WARNING]
> Note that all these changes here are still WIP. Feedback is appreciated but provide it with the knowledge that things will change 🙇

PR #410, opened 2025-02-12
feat: Add generated typed Pyroscope API client

simonswine:
This adds a Pyroscope API client, which is generated from the Pyroscope protobuf specifications.

This allows us to use a fully typed API, which evolves together with the backend's specifications.

https://connectrpc.com/docs/web/query/getting-started/
https://github.com/connectrpc/connect-query-es
https://buf.build/blog/introducing-connect-query

This is a first step towards #354.

It is also the api client we plan to use for #360 #363


simonswine:
@ifrost validated the CI fails when there are changes: https://github.com/grafana/explore-profiles/pull/424

ifrost (approved):

PR #411, opened 2025-02-12
ci: Run deployments in sequence

ifrost:
### ✨ Description

**Related issue(s):** #295

As discussed it's preferable to run steps in sequence to avoid multiple points that can fail.

### 🧪 How to test?

Not possible to test on a PR flow as it's part of the release workflow 🤷

### DIFF
--- a/.github/workflows/cut-new-release.yml
+++ b/.github/workflows/cut-new-release.yml
@@ -75,20 +75,20 @@ jobs:
     needs: [create-new-version]
     uses: ./.github/workflows/build.yml
     with:
-      ref: ${{ needs.create-new-version.outputs.version }}
+      ref: ${{ jobs.create-new-version.outputs.version }}
       e2e: false
 
   upload:
-    needs: [create-new-version, build]
+    needs: [build]
     uses: ./.github/workflows/upload.yml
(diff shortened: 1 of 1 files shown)


.github/workflows/cut-new-release.yml
```diff
@@ -75,20 +75,20 @@ jobs:
     needs: [create-new-version]
     uses: ./.github/workflows/build.yml
     with:
-      ref: ${{ needs.create-new-version.outputs.version }}
+      ref: ${{ jobs.create-new-version.outputs.version }}
```
ifrost, line 78: It won't work unfortunately. According to https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs the job needs to be listed in `need` to get access to its output.

ifrost: I had to revert it in 54936838a3b4b999db3183a1345ef7031517cdef. Tested that indeed it won't work in https://github.com/ifrost/deployme/actions/runs/13291199875

grafakus (approved):

grafakus (approved):

PR #429, opened 2025-02-19
feat(investigations): adapt to changed plugin id

svennergr:
### ✨ Description

Change pluginId of the investigations app. 
https://github.com/grafana/investigations-app/pull/175


ifrost:
Is this change already in ops/prod? Should we align when it's released or use two ids for backwards compatibility?

svennergr:
> Is this change already in ops/prod? Should we align when it's released or use two ids for backwards compatibility?

Yea, we deployed the change in investigations already. Investigations isn't released in prod anywhere, so we felt we should rather break now, before it gets troublesome after we release. So, feel free to merge without using both IDs.

ifrost (approved):

PR #445, opened 2025-02-27
feat(FlameGraph): Show the flame graph for a selected span

joey-grafana:
### ✨ Description

Allows users to jump to multiple places (split view and Profiles Drilldown) depending on their use case via the Trace View.

### 📖 Summary of the changes

Adds a link from the Trace View to the Profiles Drilldown via the `Profiles for this span button`.

### 🧪 How to test?

Build this Grafana branch: https://github.com/grafana/grafana/pull/101422 and go to the trace view, click on the `Profiles for this span` dropdown and select the `Open in Profiles Drilldown` button.

You will also need to run docker for [this tracing / Pyroscope example](https://github.com/grafana/pyroscope/tree/main/examples/tracing/tempo) to generate traces with profiles.

Then in Profiles Drilldown you can see that the query in the network tab has added the `spanSelector`, service name etc from the Trace View.

Adds the `Open in Profiles Drilldown` button.

Screenshot 2025-02-27 at 14 52 51

Adds the `Span selector added` label.

Screenshot 2025-02-27 at 14 59 39


tsconfig.json
```diff
@@ -9,6 +9,8 @@
     "noUnusedParameters": false,
     "paths": {
       "@shared/*": ["./src/shared/*"]
-    }
+    },
+    "esModuleInterop": true,
+    "jsx": "react"
```
joey-grafana, line 14: Apparently needed these for my IDE

ifrost: What errors are you getting? 🤔  These values are already[ set as defaults](https://github.com/grafana/tsconfig-grafana/blob/master/base.json), we should not need to override them.

src/shared/domain/url-params/useSpanSelectorFromUrl.ts
```diff
@@ -0,0 +1,12 @@
+import { useUrlSearchParams } from './useUrlSearchParams';
+
+export function useSpanSelectorFromUrl(): [string | undefined, (newSpanSelector: string) => void] {
```
ifrost, line 3: For handling state in URL we use Scene variables synced to the URL.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SpanSelectorLabel.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useSpanSelectorFromUrl } from '@shared/domain/url-params/useSpanSelectorFromUrl';
+import React from 'react';
+
+export function SpanSelectorLabel() {
+  const styles = useStyles2(getStyles);
+  const [spanSelector, setSpanSelector] = useSpanSelectorFromUrl();
+
+  return (
+    <div className={styles.container}>
+      <Tooltip content={`You have added a span selector to the flamegraph query (${spanSelector}).`} placement="top">
```
ifrost, line 12: It's not very clear in the UI that span selector was applied 🤔 Also using the example it didn't find the data for that span. We could be more clear about it as well. It's weird also that the link "profiles for this span" renders different result than "Open in Profiles Drilldown"

Screenshot 2025-03-03 at 15 45 10

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
```diff
@@ -171,6 +178,38 @@ export class SceneFlameGraph extends SceneObjectBase<SceneFlameGraphState> {
       }
     }, [isAiButtonDisabled, sidePanel]);
 
+    // Remove span selector when dependencies change
+    const serviceName = getSceneVariableValue(model, 'serviceName');
```
ifrost, line 182: I think SceneFlameGraph shouldn't handle this logic, it's responsibility it's to handle displaying the FlameGraph not controlling the flow of the app. We have similar logic when [exploration type changes](https://github.com/grafana/explore-profiles/blob/1c7b6c3f537b9ff0bc9060bac5f0e2c240ba00c8/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx#L285), [service name changes](https://github.com/grafana/explore-profiles/blob/1c7b6c3f537b9ff0bc9060bac5f0e2c240ba00c8/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx#L210) or when [data source changes](https://github.com/grafana/explore-profiles/blob/1c7b6c3f537b9ff0bc9060bac5f0e2c240ba00c8/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx#L202) in the root scene - [ScenesProfilesExplorer](https://github.com/grafana/explore-profiles/blob/1c7b6c3f537b9ff0bc9060bac5f0e2c240ba00c8/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx). It'd be easier to follow if it was put there 🤔

ifrost: 1. The state (URL param) is changed inside the view. This means if the component is unmounted (e.g. due to rendering error) the logic to clean up the state will not be triggered and will be gone keeping the URL stale.

2. Changes in the state should be rather handled by the scene object (check also https://grafana.com/developers/scenes/core-concepts: "We suggest that you implement the state-modifying methods in the scene object rather than in the component to separate the model complexity from the component."). This makes testing simpler.

3. Decentralized state management can cause obscure bugs. Changes to state happen inside the SceneFlameGraph - this couples SceneFlameGraph with other variables. Introducing a new variable in the future would require changing `SceneFlameGraph` accordingly if needed but it's easy to miss when adding new features.

4. Apart from that I think we should also react to changes to exploration type. Switching between views keeps the span selector which may be confusing (e.g. in the FlameGraph overview you can click "Labels" and select a different label but it will keep the span selector which may end up in no FlameGraph being rendered)


5. What's more I think the current logic is a bit coincidental:

```
const serviceName = getSceneVariableValue(model, 'serviceName');
```

getSceneVariableValue is not a React hook so it won't trigger useEffect on change. It seems to work because the component is re-rendered but it may be coming from some other properties.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SpanSelectorLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <span>Span selector added</span>
+      </Tooltip>
+      <Button
+        size="md"
+        fill="text"
+        variant="secondary"
+        icon="times"
+        tooltip={`Remove span selector from query`}
+        tooltipPlacement="top"
+        onClick={() => {}}
```
joey-grafana, line 20: Looks like we needed to use a variable for state management. But I will need to utilise your expertise here @Pitor for the best way to remove the spanSelector when this button is clicked? (not sure what way to go about this)

joey-grafana, line 20: Looks like we needed to use a variable for state management. But I will need to utilise your expertise here 
@ifrost for the best way to remove the spanSelector when this button is clicked? (not sure what way to go about this)

ifrost: This can be done with events. The lower-level component dispatches an event ([example](https://github.com/grafana/explore-profiles/blob/39e550a169323f53e8b0a4f9481f34009ea7e8f3/src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx#L130)) and the root component which control variables can listen to the event and change variable accordingly ([example](https://github.com/grafana/explore-profiles/blob/1c7b6c3f537b9ff0bc9060bac5f0e2c240ba00c8/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx#L231)).

ifrost (approved):

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/domain/events/RemoveSpanSelector.ts
```diff
@@ -0,0 +1,7 @@
+import { BusEventWithPayload } from '@grafana/data';
+
+export interface RemoveSpanSelectorPayload {}
+
+export class RemoveSpanSelector extends BusEventWithPayload<RemoveSpanSelectorPayload> {
+  public static type = 'diff-auto-select';
```
joey-grafana, line 6: nit: change type

ifrost (dismissed):

ifrost (approved):

PR #450, opened 2025-03-09
docs: Update the readme for Drilldown with proper navigation instructions

09jvilla:
Readme was mostly updated but it looks like we were still saying that this app was nested under "Explore" even though now its under "Drilldown".


ifrost (approved):
Thanks 🚀

Issue #423, opened 2025-02-14
[enhancement] Re-enable lazy loading


ifrost:
### Feature description

**Is your feature request related to a problem? Please describe.**

After enabling lazy loading of some components in https://github.com/grafana/explore-profiles/pull/322 we faced too many issues like:

<img width="963" alt="Image" src="https://github.com/user-attachments/assets/33cc4d72-cc32-4ce9-883f-9d7c392f7e66" />

With the error:

```
explore:1 Failed to find a valid digest in the 'integrity' attribute for resource 'https://(...)/public/plugins/grafana-pyroscope-app/75.js' with computed SHA-256 integrity 'RBRryzBVsbMkhV7se/rmF2da0ND4qBStJhaK0HsTGbE='. The resource has been blocked.
```

Also seen this:

```
Refused to execute script from 'https://(...).net/public/plugins/grafana-pyroscope-app/75.js' because its MIME type ('application/json') is not executable, and strict MIME type checking is enabled.
```

Disabling browser cache helps to mitigate suggesting it might be due to caching problems.

See also [this Slack thread](https://raintank-corp.slack.com/archives/C077S1MP40M/p1739884656115289)

**Describe the solution you'd like**

Components should be lazy-loaded and no error should appear.

We may consider:

- adding cache busting to the resources
- checking caching policies / configuration
- adding `[contenthash]` to produced files

PR #455, opened 2025-03-11
feat(CI): Append SHA to version number when releasing from main

ifrost:
### ✨ Description

Fixes: #470
Related issue(s): #423 

### 📖 Summary of the changes

This is to append short commit sha to non-tagged builds: in main or in PRs. This way we can add a cache busting to the app (note `_cache` param passed in the URL)

Screenshot 2025-03-11 at 14 27 26

This is causing problems with getting main / PRs out to dev and ops environments. Without clearing the cache customers are stuck with the cached version even after instance restart.

This will be needed to properly test #423 

### 🧪 How to test?

You can upload the version to your instance and should see the request with correct _cache and version number in:

Screenshot 2025-03-11 at 14 27 12


joey-grafana:
| You can upload the version to your instance and should see the request with correct _cache and version number in:

What terminal commands do you use to upload a particular version? Also, where do you get the exact version number/url?

ifrost:
> | You can upload the version to your instance and should see the request with correct _cache and version number in:
> What terminal commands do you use to upload a particular version? Also, where do you get the exact version number/url?

Just using regular commands to set the version of the provisioned plugin on an instance. You can pass `version=75ffc8fc3065fdfce309dcf51d9f93d6ecabf2db` as the version.

The `_cache` param is added to `grafana-pyroscope-app/module.js` - you should see it in the network tab when loading Grafana on your instance. The version is passed in the URL, you can also check the version in the top right corner of the app in "info" pop up (see screenshots in the PR description)

joey-grafana (approved):
LGTM! 👍 

Screenshot 2025-04-02 at 11 56 48

PR #457, opened 2025-03-11
chore: Use label to upload artifacts

ifrost:
### ✨ Description

Fixes: 
Related issue(s):

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,11 @@ on:
       - docs/sources/**
     branches:
       - main
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
 
 env:
   BUNDLEWATCH_GITHUB_TOKEN: ${{secrets.BUNDLEWATCH_GITHUB_TOKEN}}
@@ -31,11 +36,12 @@ jobs:
   upload:
     name: Upload to GCS
     needs: [build]
+    if: github.event.label.name == 'upload'
     uses: ./.github/workflows/upload.yml
     secrets: inherit
     with:
       version: ${{ github.event.pull_request.head.sha || github.sha }}
-      github_environment: ${{ github.event_name != 'push' && 'gcs' || 'gcs-no-approval' }}
+      github_environment: gcs-no-approval
 
   deploy-main-to-dev-and-ops:
     name: Deploy


.github/workflows/ci.yml
```diff
@@ -10,6 +10,11 @@ on:
       - docs/sources/**
     branches:
       - main
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
```
joey-grafana, line 17: Was this being run for too many types previously?

joey-grafana (commented):
Can you please elaborate on

> [!IMPORTANT]
> Before merging to main update restriction rules for gcs-no-approval

.github/workflows/ci.yml
ifrost: No, we're adding "labeled", but to do this we also need to include all defaults for "types"

.github/workflows/ci.yml
```diff
@@ -31,11 +36,12 @@ jobs:
   upload:
     name: Upload to GCS
     needs: [build]
+    if: github.event_name == 'push'
     uses: ./.github/workflows/upload.yml
     secrets: inherit
     with:
       version: ${{ github.event.pull_request.head.sha || github.sha }}
-      github_environment: ${{ github.event_name != 'push' && 'gcs' || 'gcs-no-approval' }}
+      github_environment: gcs-no-approval
```
ifrost, line 44: cc @joey-grafana

> Can you please elaborate on

> [!IMPORTANT]
> Before merging to main update restriction rules for gcs-no-approval 

Before this change we used:
- "gcs" GitHub environment for pushes to branches (configured only for PRs)
- "gcs-no-approval" for main and tags (so it runs automatically)

With this change we will use "gcs-no-approval" for all, including PRs

However "gcs-no-approval" has a restriction rule that it can be applied only on "main" and "v*" branches. When we merge this PR we can remove the restriction.

Anyone who can add a label will be able to upload the artifact = anyone who can create a PR in the main repo (not forks)

joey-grafana: For context: spoke with Piotr who mentioned only Grafana users can trigger an upload from the label 👍

joey-grafana (approved):

PR #459, opened 2025-03-11
fix(FlameGraph): Use entire file as context in "Optimize Code" prompt

bryanhuhta:
### ✨ Description

Currently, we only use the snippet of code displayed to the user as context for the LLM prompt. This has the downside of making the LLM reason with less information. If we pass the entire file (if we have it) to the LLM prompt, this should improve its reasoning ability and give users the chance to ask follow up questions about portions of the code which might be in the same file, but not in the profiling snippet.

Here's an example of a short code snippet in a large file:

image-1

When asked about code in a different part of the same file, the LLM reports it can only reason about the current snippet:

image

With the changes in this PR, the LLM can now reason about the the entire file. Here's a snippet that's part of a larger file:

Screenshot 2025-03-11 at 5 17 43 PM

And here is the LLM's response to asking about a specific part of the file not found in the snippet:

Screenshot 2025-03-11 at 5 07 43 PM

### 🧪 How to test?

The unit tests should pass

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/CodeContainer/domain/buildLineProfiles.ts
+++ b/src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/CodeContainer/domain/buildLineProfiles.ts
@@ -4,9 +4,9 @@ const VERTICAL_LINES_PADDING = 5;
 
 type CallSitesMap = Map<number, CallSiteProps>;
 
-export function buildPlaceholderLineProfiles(callSitesMap: CallSitesMap) {
+export function annotatePlaceholderLineProfiles(callSitesMap: CallSitesMap): LineProfile[][] {
   if (!callSitesMap.size) {
-    return [];
+    return [[], []];
   }
 
   const callSites = Array.from(callSitesMap.values()).sort((a, b) => a.line - b.line);
@@ -14,12 +14,11 @@ export function buildPlaceholderLineProfiles(callSitesMap: CallSitesMap) {
   const firstLineIndex = Math.max(0, callSites[0].line - VERTICAL_LINES_PADDING - 1);
   const lastLineIndex = callSites[callSites.length - 1].line + VERTICAL_LINES_PADDING + 1;
 
-  const lines = [];
-
+  const annotatedSnippet = [];
   for (let lineNumber = firstLineIndex + 1; lineNumber < lastLineIndex; lineNumber++) {
     const callSite = callSitesMap.get(lineNumber);
 
-    lines.push({
+    annotatedSnippet.push({
       line: undefined,
       number: lineNumber,
       cum: callSite?.cum ?? 0,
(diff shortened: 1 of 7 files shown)


src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/CodeContainer/domain/buildLineProfiles.ts
```diff
@@ -4,45 +4,39 @@ const VERTICAL_LINES_PADDING = 5;
...
+  // the appropriate line numbers, but no content.
+  return [annotatedSnippet, []];
 }
 
-export function buildLineProfiles(fileContent: string, callSitesMap: CallSitesMap): LineProfile[] {
-  if (!callSitesMap.size) {
-    return [];
-  }
-
+export function annotateLines(fileContent: string, callSitesMap: CallSitesMap): LineProfile[][] {
```
ifrost, line 34: It'd be good to change the name of the file as well to avoid confusion (or keep the old name)

ifrost, line 34: Could we return a struct/map instead of nested array for better readability? e.g.

```
{ 
  snippetLines: LineProfile[],
  allLines: LineProfile[] 
}
```

instead of `LineProfile[][]`

ifrost (commented):
Looks great!

We're passing all lines now but annotate usage only the snippet, and other lines would have e.g. cpu time set to `0` here:

```
    return {
      line,
      number: lineNumber,
      cum: callSite?.cum ?? 0,
      flat: callSite?.flat ?? 0,
    };
``` 

Could that confuse the AI indicating that everything else in the file is super fast? 🤔 

example:

Screenshot 2025-03-12 at 11 55 14

bryanhuhta:
> Could that confuse the AI indicating that everything else in the file is super fast? 🤔

Interesting observation. From my testing I personally haven't seen this cause a lot of problems. However, I think it's worth keeping an eye out for.

We could give the LLM extra information to ignore the performance value of certain lines, but this will require the prompt to be changed. For example, for lines with not profiling information, we could render a `-` instead of a `0`, but we'd need to update the prompt to give the LLM this knowledge. I'd like to see such a change occur in its own PR for debugability.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/CodeContainer/domain/buildLineProfiles.ts
bryanhuhta: 👍 fixed in https://github.com/grafana/explore-profiles/pull/459/commits/071a438d6955ee5c20092150789d20c5bb8a26f9.

bryanhuhta: 👍 fixed in https://github.com/grafana/explore-profiles/pull/459/commits/071a438d6955ee5c20092150789d20c5bb8a26f9

ifrost (approved):

PR #462, opened 2025-03-13
feat(Metrics): Basic UI for experimental metrics from profiles

ifrost:
### ✨ Description

Fixes: #464
Related issue(s): https://github.com/grafana/pyroscope-squad/issues/340

### 📖 Summary of the changes

Built on top of https://github.com/grafana/explore-profiles/pull/360 with following changes:

Misc:
- The feature needs to be enabled in settings. We can use it to get it out quicker and iterate more frequently more easily.
- RecordingRules API is used (with protobuf types)

Create a recording rule flow:
- The state of the modal is reset each time it's opened instead of showing stale inputs
- Create a recording rule with no labels is fixed 
- Small UI/UX adjustments in the modal:
  - Inlining alert icon and text in the validation
  - Showing validation errors when input changes to give quick feedback
  - Using plain text instead of chilclets (simpler in code and won't create an impression it's editable)
 
List recording rules view:
- Added back button (also added to Ad Hoc view for consistency)
- Using InteractiveTable to simplify the code
- Deleting needs additional confirmation
- List is refreshed after deleting
- Lazy loading is disabled (we need to tackle it separately in #423)
- Error / empty state page was added

What's not included?
- Validation when a metric with the same name is created (it overrides existing metric)
- There's no way to open a recoding rule in Explore/Metrics Drilldown (the user would need to go there manually)
- No additional user disclaimer/consent screen (to be done in https://github.com/grafana/explore-profiles/issues/463)
- Not much user tracking (only selecting "view recording rules" is being tracked)
- To be followed up in https://github.com/grafana/explore-profiles/issues/466

### 🧪 How to test?

Make sure you use latest weekly build from Pyroscope with recording_rules feature toggle enabled.

- Enable "metrics from profiles" in Settings menu
- Open a flame graph
- Click on the top "total" block
- Select "Create recording rule"
- Enter a metric name, select labels
- Click "Create"
- Open "View recording rules" from the top right corner
- List recording rules
- Try deleting a recording rule

https://github.com/user-attachments/assets/1f3d7f18-ee82-498c-aa7a-dab0ea2e289a


simonswine:
Noticed that the menu button jumps from Profiles to Explore once I go to setting/recording rules:

Screenshot 2025-03-27 at 17 39 23

Screenshot 2025-03-27 at 17 39 29

simonswine (approved):
LGTM

I think it generally works for now and is good enough.

But it is hard how the feature can be found. I also think that some valid use cases are not well supported: Let's say I have 100 services and I want to enable recording rules for all of them.

alsoba13:
Looks great!

I agree with Christian, I think we could iterate later to make the recording rule creation more visible and flexible. Let's get this merged and we can move this discussion somewhere else.

alsoba13 (approved):

PR #465, opened 2025-03-18
feat(investigations): add `fieldConfig` to investigations context

svennergr:
### ✨ Description

Corresponding PR for https://github.com/grafana/investigations-app/pull/229 to pass `fieldConfig` to the investigations app in order to keep same colors, etc of panels.


src/pages/ProfilesExplorerView/components/SceneLabelValuesTimeseries/SceneTimeseriesMenu.tsx
```diff
@@ -140,7 +141,53 @@ export class SceneTimeseriesMenu extends SceneObjectBase<SceneTimeseriesMenuStat
       ) as TimeSeriesQuery;
   }
 
+  getFieldConfig() {
```
ifrost, line 144: Isn't it the same as https://github.com/grafana/traces-drilldown/pull/369? Could we just pass the data / fields and the investigation app would calculate the config?

svennergr: It is the same, yes. I wanted to avoid passing super large objects, which `data` could obviously be. Those objects are all formed into a Proxy and that makes working with those objects a bit more difficult. 

With a broader vision, we should really move the whole component into the investigations app and then use it in Drilldown apps with exposed components. That would simplify things for the Drilldown integrations.

svennergr: Do you think we could merge this never the less like it is?

ifrost: > It is the same, yes. I wanted to avoid passing super large objects, which data could obviously be.

What's the problem with passing big objects? 🤔  wouldn't it be actually easier if anyway we want to have a component inside investigations app?

> Do you think we could merge this never the less like it is?

If I need to maintain this code I need to double check again why we need these logic, it's not super obvious from reading the code. And we will probably need some tests to see a few examples.

ifrost:
Hey @svennergr 👋 I'll move https://github.com/grafana/profiles-drilldown/pull/465 and https://github.com/grafana/traces-drilldown/pull/369 back to draft. Please re-request the review when it's ready!

Issue #468, opened 2025-03-20
chore: Upgrade grafana/llm


ifrost:
We need to upgrade llm to at least 0.12.0 in order to stop depending on uplot 1.6.30

See also https://raintank-corp.slack.com/archives/C06J391UD8B/p1740500313657659

PR #469, opened 2025-03-20
chore: Upgrade llm package

ifrost:
### ✨ Description

Fixes: #468

### 📖 Summary of the changes

Upgrading to 0.12.0. Tried upgrading to 0.14.0 but the explain functions stopped working without any errors - needs a separate investigation

### 🧪 How to test?

Explain Flame Graph should work as expected


joey-grafana (approved):
Installed the new package, built the plugin, reran dev env and the explain seems to be working well 👍

Issue #467, opened 2025-03-19
[bug] Incorrect settings are loaded when data source changes


ifrost:
### Bug description

- **Grafana version:** 11.5.0
- **Plugin version:** 1.2.0

### Steps to reproduce

1. Create at least 2 data sources to Pyroscope
2. Open Profiles Drilldown
3. Open settings
4. Check Network tab if the URL to get settings contains the correct data source UID
5. Go back to the app
6. Change the data source
7. Open settings
8. Check Network tab if the URL to get settings contains the correct data source UID

#### Actual behaviour

- The data source uid is never updated, it uses the first used only

#### Expected behaviour

- The data source uid should be updated

It also loads incorrect settings for Flame Graph view when data source is switched

PR #474, opened 2025-03-27
fix(Settings): Reload settings when data source changes

ifrost:
### ✨ Description

Fixes: #467

### 📖 Summary of the changes

Ensure base URL is recreated for settings when fetching + when data source changes the query is retriggered.

### 🧪 How to test?

Looks for `/api/datasources/proxy/uid/[DATASOURCE_UID]/settings.v1.SettingsService/Get` calls in network tab. The `[DATASOURCE_UID]` should be updated after changing the data source.

Before:

https://github.com/user-attachments/assets/c3ee7581-56cf-409c-afdf-be5d1061dcbb

After:

https://github.com/user-attachments/assets/0ec49778-20f8-4462-8cba-4fd54429ac83

### DIFF
--- a/src/shared/infrastructure/settings/settingsApiClient.ts
+++ b/src/shared/infrastructure/settings/settingsApiClient.ts
@@ -38,4 +38,6 @@ class SettingsApiClient extends ApiClient {
   }
 }
 
-export const settingsApiClient = new SettingsApiClient();
+export const useSettingsApiClient = () => {
+  return new SettingsApiClient();
+};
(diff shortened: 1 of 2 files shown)


alsoba13:
This is great! Apart from this, I believe we have some kind of bug in the backend where we wipe the settings on deployment

alsoba13 (approved):

joey-grafana (approved):
Nice, works as expected! 👍

PR #475, opened 2025-04-01
WIP: Provide extension point for other plugin to add a settings tab.

simonswine:
TODO

### DIFF
--- a/src/pages/SettingsView/SettingsView.tsx
+++ b/src/pages/SettingsView/SettingsView.tsx
@@ -1,16 +1,29 @@
 import { css } from '@emotion/css';
 import { GrafanaTheme2 } from '@grafana/data';
-import { Button, FieldSet, InlineField, InlineFieldRow, InlineSwitch, Input, useStyles2 } from '@grafana/ui';
+import {
+  Button,
+  FieldSet,
+  InlineField,
+  InlineFieldRow,
+  InlineSwitch,
+  Input,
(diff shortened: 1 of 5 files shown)


src/pages/SettingsView/SettingsView.tsx
```diff
@@ -26,71 +39,110 @@ export default function SettingsView() {
...
+        </FieldSet>
+      ),
+    },
+  ];
+
+  const { components, isLoading } = useSettingsExtensions();
+  const pyroscopeDataSource = ApiClient.selectDefaultDataSource();
+  const pluginTabs = components.map((Component) => {
+    return {
+      label: (Component as unknown as PluginExtensionComponent).title || 'Unknown Extension',
```
simonswine, line 110: @jarben a bit unsure, how to get that title properly.

I do think the types are exported in @grafana 11.6 and later, but I couldn't make the upgrade work.

jarben: Do you need to access it via component.meta? Similarly to this: https://github.com/grafana/grafana/blob/608a2a75358e041bcb69a849cbd3072bccd949e4/public/app/features/datasources/components/EditDataSource.tsx#L251

simonswine: I think I can't because those types are not exported. (At least not in v11.5.0)

PR #447, opened 2025-02-28
feat: "profiling insights" extension

darrenjaneczek:
### 📖 Summary of the changes

Provide an extension point for a "insights" button.

### 🧪 How to test?

See https://github.com/grafana/hackathon-2025-03-profiling-insights

Under `./plugins/grafana-insights-app/README.md` there will be some descriptions of how to run that plugin and attach it to this plugin (using this branch).

image

PR #477, opened 2025-04-02
fix: Revert "feat: "profiling insights" extension (#447)"

ifrost:
This reverts commit 13ff552083cc4555aa21e8401be1516a364b8271.

### ✨ Description

Fixes: #447 

Reverting insights extension as the error is coming from the plugin. We can re-enable it once it has better support for error handling.


simonswine (approved):
LGTM

PR #484, opened 2025-04-09
fix: Cascader options break when services have multiple levels with matching prefix

simonswine:
Assuming there is an overlap of services on different levels in the service_name hierachie, the app crashses.

image

I created that example by having two profiles series:

- `my-service`
- `my-service/my-subservice`

I have reimplemented the function sorting the services names first and then building up a hierarchic map, which should also be faster.

What is not ideal yet, is that it is not clear which one of the two is the one with children vs. the main one:

image
image


ifrost (approved):
Thanks for fixing it! I agree it's not ideal but still better than an error 😅

PR #487, opened 2025-04-15
fix: Correct the menu for explore-profiles

### LINKED ISSUE #486: [bug] Menu behaves wrong with subpages in Drilldown Profiles app
### Bug description

We seem to have some bug in how we've configured the menu:

* When I go to the sub pages like adhoc and settings, the "Drilldown" gets highlighted instead of "Drilldown -> Profiles"

### Steps to reproduce

See video (recorded in ops with v1.2.3 and v12.0.0-86517):

https://github.com/user-attachments/assets/79e61221-b8ad-43ad-ae7c-f0d8201a848c

simonswine:
This is my first attempt to fix #486, but it might have some backwards compatibility problems, as `/explore` goes away.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -5,7 +5,7 @@ export const PYROSCOPE_APP_ID = plugin.id;
 export const PLUGIN_BASE_URL = `/a/${PYROSCOPE_APP_ID}`;
 
 export enum ROUTES {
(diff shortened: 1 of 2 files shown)


ifrost:
Closing this for now to clean up the board. I'll need to have another look at it. Ideally we do not want to change the URL

PR #490, opened 2025-04-16
feat: visualize profile annotations

aleks-p:
### ✨ Description

Time series from profiling data can experience drops/increases when ingest throttling is enabled/disabled. To provide more context to users on when and why this happens, the Pyroscope API [now provides annotations as part of the series data](https://github.com/grafana/pyroscope/pull/3956).

This change visualizes annotations present in series responses, within time series panels.

Screenshot 2025-04-16 at 14 18 19

Ideally I would like to make it possible for users to disable these annotations if they find them distracting. This information would get stored in local storage (like Favorites). I will look into ways we can do this in the UI.

### 📖 Summary of the changes

We subscribe to the time series panel data to extract and process annotations optionally present in the data. Annotations come as a separate data frame (see https://github.com/grafana/grafana/pull/104130 for more info).

The displayed annotations are range-based when the start and end of the "event" is within the range of the time series panel. When the "end" time of the event is in the future, we render non-ranged (regular) annotations.

### 🧪 How to test?

I've added tests for the annotation processing part. For a full end-to-end test we need:

- Grafana with https://github.com/grafana/grafana/pull/104130
- A recent version of Pyroscope (including https://github.com/grafana/pyroscope/pull/3956)
- A way to apply ingestion limits (throttling) to profiles, e.g., by providing tenant overrides for Pyroscope
  ```
    overrides:
      anonymous:
        ingestion_rate_mb: 200
        ingestion_limit:
          period_type: hour
          period_limit_mb: 128
          limit_reached: true
          limit_reset_time: 1744829133
          sampling:
            num_requests: 1
            period: 10s
        max_profile_stacktrace_samples: 32000
  ```

I will try to make testing easier once the Grafana change gets merged.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneMainServiceTimeseries.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneMainServiceTimeseries.tsx
@@ -17,7 +17,7 @@ import { FiltersVariable } from '../domain/variables/FiltersVariable/FiltersVari
 import { GroupByVariable } from '../domain/variables/GroupByVariable/GroupByVariable';
 import { ProfileMetricVariable } from '../domain/variables/ProfileMetricVariable';
 import { ServiceNameVariable } from '../domain/variables/ServiceNameVariable/ServiceNameVariable';
-import { createAnnotationFrame } from '../helpers/annotationCollector';
+import { createThrottlingAnnotationFrame } from '../helpers/createThrottlingAnnotationFrame';
 import { getSceneVariableValue } from '../helpers/getSceneVariableValue';
 import { PYROSCOPE_DATA_SOURCE } from '../infrastructure/pyroscope-data-sources';
 import { getProfileMetricLabel } from '../infrastructure/series/helpers/getProfileMetricLabel';
@@ -158,7 +158,7 @@ export class SceneMainServiceTimeseries extends SceneObjectBase<SceneMainService
         // add annotation for the first time
         if (!newState.data.annotations?.length && !prevState.data?.annotations?.length) {
           const { $data } = this.getTimeseries()!.state;
-          const annotationFrame = createAnnotationFrame(newState.data);
+          const annotationFrame = createThrottlingAnnotationFrame(newState.data);
 
           $data?.setState({
             data: {
--- a/src/pages/ProfilesExplorerView/components/SceneMainServiceTimeseries.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneMainServiceTimeseries.tsx
@@ -1,3 +1,4 @@
+import { LoadingState } from '@grafana/data';
 import {
   SceneComponentProps,
   SceneDataTransformer,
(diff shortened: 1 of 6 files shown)


src/pages/ProfilesExplorerView/components/SceneMainServiceTimeseries.tsx
```diff
@@ -143,6 +148,39 @@ export class SceneMainServiceTimeseries extends SceneObjectBase<SceneMainService
...
+    this._subs.add(
+      this.getTimeseries()?.state.$data?.subscribeToState((newState, prevState) => {
+        if (!newState.data || newState.data.state !== LoadingState.Done) {
+          return;
+        }
+
+        // add annotation for the first time
+        if (!prevState.data?.annotations?.length) {
+          const { $data } = this.getTimeseries()!.state;
+          const annotationFrame = createThrottlingAnnotationFrame(newState.data);
```
ifrost, line 161: If I read this correctly we transform annotations frame into a format used by TimeSeries visualization. Could we do it in [data source plugin code](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/grafana-pyroscope-datasource/datasource.ts#L32)? Or even in the backend as part of [this PR](https://github.com/grafana/grafana/pull/104130)? This way we could get it supported more natively and it would work automatically for all time series panels and even in Grafana Dashboards.

aleks-p: The reason I opted to do it here was to allow the frontend to decide how these annotations are visualized.

Annotations are serialized in json, and can contain arbitrary fields. We anticipate that we can have other annotations in the future, e.g., automatically detected service version changes, horizontal scale in/out events, user annotations, etc. I don't think we would want to force the components in the middle to have to understand and transform very specific messages, but let me know if I've missed something.

ifrost: > (...) allow the frontend to decide how these annotations are visualized.

Is this annotation meaningful only in `SceneMainServiceTimeseries.tsx` and only in Profiles Drilldown app? If so maybe annotations should not be returned with every metric query? 

I feel like it'd be useful to have this annotation in Dashboards and in Explore as well. The format used by Pyroscope API is absolutely fine, it should be anything you find most useful, not bound to the app.

If we moved transforming Pyroscope API annotations to Grafana Annotations to DatasourceApi.query (or in plugin's backend code) then Explore and Dashboards could benefit from it :)

aleks-p: Thanks for your feedback @ifrost. I have moved the transformation to the datasource backend in the latest update to https://github.com/grafana/grafana/pull/104130. This works and as you said annotations now appear in any place we render time series.

The downside is that we have less control. E.g., if we wanted to give drilldown users an option to disable annotations because they are noisy, this will be harder. If we want to visualize annotations in a different way (e.g., show them in a table) we would need to parse them again, in a different place.

To be clear, I am ok with closing this right now in favour of doing everything in https://github.com/grafana/grafana/pull/104130 - but wanted to provide extra context on the original intent for how this would work.

PR #504, opened 2025-04-25
chore(ci): Fix uploading artifacts

ifrost:
### ✨ Description

Fixes issues when uploading https://github.com/grafana/profiles-drilldown/actions/runs/14659481025/job/41140864301

Upload GitHub action uploads to GCS + GitHub artifacts, but we cannot create artifacts with the same name. The GitHub artifact is now used only for creating a GitHub release so we can actually run it only when making a new release.


joey-grafana (approved):
Makes sense! But can't really test so will approve and let you test :D

PR #522, opened 2025-05-19
feat(metrics): Show annotations

ifrost:
### ✨ Description

Fixes: 
Related issue(s):

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneLabelValuesTimeseries/SceneLabelValuesTimeseries.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneLabelValuesTimeseries/SceneLabelValuesTimeseries.tsx
@@ -65,7 +65,8 @@ export class SceneLabelValuesTimeseries extends SceneObjectBase<SceneLabelValues
             new SceneDataTransformer({
               $data: buildTimeSeriesQueryRunner(
                 item.queryRunnerParams,
-                displayAllValues ? undefined : LabelsDataSource.MAX_TIMESERIES_LABEL_VALUES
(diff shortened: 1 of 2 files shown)


src/pages/ProfilesExplorerView/components/SceneLabelValuesTimeseries/SceneLabelValuesTimeseries.tsx
```diff
@@ -65,7 +65,8 @@ export class SceneLabelValuesTimeseries extends SceneObjectBase<SceneLabelValues
             new SceneDataTransformer({
               $data: buildTimeSeriesQueryRunner(
                 item.queryRunnerParams,
-                displayAllValues ? undefined : LabelsDataSource.MAX_TIMESERIES_LABEL_VALUES
+                displayAllValues ? undefined : LabelsDataSource.MAX_TIMESERIES_LABEL_VALUES,
+                true
```
aleks-p, line 69: In the future we'd perhaps want this to be a client-side (browser storage) setting, but right now it is perfect as it still allows us to build a patch with it disabled if needed.

aleks-p (approved):
Looks good to me!

src/pages/ProfilesExplorerView/components/SceneLabelValuesTimeseries/SceneLabelValuesTimeseries.tsx
aleks-p: Perhaps both? The settings page can hold the general preference, the graph level toggle can be used for ad-hoc changes. 

Keep in mind that the settings page is currently used for tenant-wide settings so it might be confusing to throw in a user (browser) level setting in there.

PR #524, opened 2025-05-22
chore(ci): Stop using repo secrets

ifrost:
### ✨ Description

This is to simplify our config as we do not need to use GitHub repo secrets


joey-grafana (approved):
We will need to trigger the workflows to properly test it's 100% good but the PR LGTM! 👍

Issue #497, opened 2025-04-23
[Metrics from profiles] Create a recording rule from a function name in the UI


ifrost:
Profiles Drilldown changes required to support https://github.com/grafana/pyroscope-squad/issues/340

More info TBC

PR #547, opened 2025-07-15
chore: Update API type definitions

ifrost:
### ✨ Description

Updating to latest type definitions. I new optional property to allow passing function names was added. It will be used in #497


ifrost:
> not 100% sure how to appropriately test / check

Thanks. These are just new types, not in use yet but required for https://github.com/grafana/profiles-drilldown/pull/543 (split the PR to make it easier to review)

PR #553, opened 2025-07-22
ifrost/fix-cookie-sync

### LINKED ISSUE #552: [bug] GitHub authentication session still shorter lived than expec
### Bug description

- **Grafana version:** v.12.1.0-16426244399
- **Plugin version:** v.1.5.0

The github cookie (created by "connect to github", with validity of 6 months), somehow reduces its validity after relaunching the plugin to a expiry "session" cookie. 

This leads to repeated logins being required, but less than before https://github.com/grafana/profiles-drilldown/pull/535.

Also on the backend, we don't see any calls to the  GithubRefresh endpoint.

### Steps to reproduce

1. Go to "function details" of a flamegraph with source repository labels
2. Connect to github
3. Check cookie `pyroscope_github_session` and its lifetime
4. Now navigate to another website e.g. https://google.com
5. Open the root of the grafana instance again (you can still see the cookiew with a long lifetime in the cookie store
6. Now open drilldown to cookie now becomes of "sesssion" expiry  and vanishes after the next leaving of the page

### Expected behaviour

Stay logged in as long as the refresh_token is valid (6 months in my case)

Note: GL interally I have also have a video of this: https://raintank-corp.slack.com/archives/C048VSWPZ4P/p1753191080963539?thread_ts=1753171397.287949&cid=C048VSWPZ4P

ifrost:
Fixes #552

When Profiles Drilldown app is initially loaded, it reads the cookie from the browser to save its value in the memory. Inadvertently it also call again `document.cookie=` with the cookie already set in the browser. However when reading a cookie with `document.cookie` the expiration time is not readable so setting the cookie to based on value from `document.cookie` resets expiration.

We can probably simplify it a bit by reading the initial cookie from the browser only in the contractor and stop syncing it from the browser in the getter. The drawback is that after deleting the cookie the user needs to refresh the browser stop using it.

Alternatively we can still sync with the browser but add additional expiration based on decoded value.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
+++ b/src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneFunctionDetailsPanel/components/GitHubContextProvider/infrastructure/GitSessionCookieManager.ts
@@ -15,13 +15,17 @@ type Cookie = {
 };
 
 class InternalGitSessionCookieManager implements GitSessionCookieManager {
-  private rawCookie: Cookie | undefined;
   private sessionCookie: GitSessionCookie | undefined;
 
+  constructor() {
+    const cookie = InternalGitSessionCookieManager.getCookieFromJar(document.cookie, GITHUB_SESSION_COOKIE_NAME);
+
+    if (cookie) {
+      this.sessionCookie = GitSessionCookie.decode(cookie.value);
+    }
+  }
+
   getCookie(): GitSessionCookie | undefined {
-    // To make sure we're using a cookie that accurately reflects the browser
-    // state, let's be paranoid and make sure our cached cookie is accurate.
-    this.syncCookieWithBrowser();
     return this.sessionCookie;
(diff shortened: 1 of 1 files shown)


simonswine:
Nice find!

> Alternatively we can still sync with the browser but add additional expiration based on decoded value.

We would have to add it to the value, as currently the decoded value has only the expiry of the access token (8 hours), but not the one of the refresh token (6 months), while the cookie gets set with the 6 months lifetime. 

> The drawback is that after deleting the cookie the user needs to refresh the browser stop using it.

Do you mind expanding a bit on that? Not super clear what delete the cookie (Drilldown itself or the user)?

ifrost:
> > The drawback is that after deleting the cookie the user needs to refresh the browser stop using it.

> Do you mind expanding a bit on that? Not super clear what delete the cookie (Drilldown itself or the user)?

If you would like to log out / stop using the current auth - the only way at the moment is to delete the git_pyroscope cookie from your browser. However with the change I made the Drilldown app will still think the cookie is there and instead of showing "connect to repo" button will show an active button to click and will try to load the source code (and fail).

> We would have to add it to the value (...)

So probably might be good to include https://github.com/grafana/profiles-drilldown/issues/187

PR #558, opened 2025-08-01
feat(AI): Integrate with Grafana Assistant

ifrost:
Related to https://github.com/grafana/grafana/pull/108684

This is start using Grafana Assistant for Flame Graph explanation instead of LLM plugin. The feature can be controlled by the feature toggles (enabled by default)

| Flame graph | Assistant integration |
|---|---|
| Single View | Use Assistant over LLM integration if available |
| Diff View | We don't have query to run in Assistant so we use LLM only |
| Ad Hoc | There's no LLM integrated because it would require formatting the data. Assistant also won't work because we don't have the query |

### DIFF
--- a/.github/workflows/plugins-pr.yml
+++ b/.github/workflows/plugins-pr.yml
@@ -1,0 +1,39 @@
+name: (New) Plugins - Publish Branch To Dev
+run-name: Deploy ${{ inputs.branch }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: Branch to publish from. Can be used to deploy PRs to dev
+        required: true
+      docs-only:
+        description: Only publish docs, do not publish the plugin
+        default: false
+        type: boolean
+
+# No global permissions, each job defines own scope
+permissions: {}
+
+jobs:
+  cd:
+    name: CI / CD
+    uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@main
+    # These are maximum permissions for the job and should match the permissions defined in the workflow
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
+    with:
+      branch: ${{ inputs.branch }}
+      environment: dev
+      # No grafana-cloud-deployment-type defined -> plugin will be published to the catalog, but not deployed to Grafana Cloud
+
+      run-playwright: false
+      run-playwright-docker: true
+      run-playwright-with-skip-grafana-dev-image: true
+      upload-playwright-artifacts: true
+      run-playwright-with-grafana-dependency: '>=12.1.0'
+      playwright-config: e2e/config/playwright.config.ci.ts
+      playwright-docker-compose-file: docker-compose.e2e.yaml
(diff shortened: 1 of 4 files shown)


docker-compose.local.yaml
```diff
@@ -5,16 +5,14 @@ services:
       context: ./.config
       args:
         grafana_image: ${GRAFANA_IMAGE:-grafana-enterprise}
-        grafana_version: ${GRAFANA_VERSION:-12.0.0}
+        grafana_version: ${GRAFANA_VERSION:-12.1.0}
```
ifrost, line 8: This is to align with changes made to e2e tests in previous PRs

package.json
```diff
@@ -42,7 +42,7 @@
     "@emotion/css": "11.10.6",
     "@grafana/data": "^12.1.0",
     "@grafana/faro-web-sdk": "^1.10.0",
-    "@grafana/flamegraph": "12.1.0",
+    "@grafana/flamegraph": "12.2.0-17195718647",
```
joey-grafana, line 45: I presume this specific version is ok to use, but have not tested it

src/shared/infrastructure/settings/featureToggles.ts
```diff
@@ -3,4 +3,6 @@ import { config } from '@grafana/runtime';
 export const featureToggles = {
   // @ts-ignore Remove after upgrading grafana/runtime to version including the toggle type
   metricsFromProfiles: !!config.featureToggles['metricsFromProfiles'],
+  // @ts-ignore Remove after upgrading grafana/runtime to version including the toggle type
```
joey-grafana, line 6: Add an issue to the board I'd say if not already done so we don't forget

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
```diff
@@ -191,6 +192,14 @@ export class SceneFlameGraph extends SceneObjectBase<SceneFlameGraphState> {
       setRecordingRulesModalState({ isOpen: true, functionName });
     });
 
+    // Do not show AI button if the assistant integration is enabled to avoid having two AI buttons in the UI
+    // For debugging purposes and comparing both you can use localStorage flag grafana-pyroscope-app.forceShowAiButton
```
joey-grafana, line 196: This feels strange to add debugging flags that enabled functionality. Do we use this pattern elsewhere in Grafana? Perhaps we can remove if it's served its purpose? Technically, users can see the PR / code and enable it themselves which may lead to strange behaviour

ifrost: There's no harm if someone really wants to show both and uses the code from the PR. This is only to be able to compare both results. Don't think there's a similar pattern anywhere - this case is a bit special because we're replacing existing AI solution with a different one.

PR #570, opened 2025-08-13
ci: Push build to ops

adrapereira:
### ✨ Description

Fixes: 
Related issue(s):

### DIFF
--- a/.github/workflows/plugins-ci.yml
+++ b/.github/workflows/plugins-ci.yml
@@ -23,7 +23,7 @@ jobs:
       branch: ${{ github.event_name == 'push' && github.ref_name || github.ref }}
 
       # When pushing to "main", publish and deploy to "dev" (CD). For PRs, skip publishing and deploying (run CI only)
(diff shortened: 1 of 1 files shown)


joey-grafana:
Probably want to do this first?
`Ensure that you have no stack overrides in ops / staging for your plugin`

Also I see the following two are not marked as complete in the [epic](https://github.com/grafana/profiles-drilldown/issues/540) fyi in case they should be done first. The first ones subtasks are all done so maybe just need to mark done
Screenshot 2025-08-13 at 09 37 20

ifrost:
This one is a separate but will be helpful to have it. It's to make sure we have Profiles Drilldown page in settings.

- [ ] https://github.com/grafana/profiles-drilldown/issues/563

This one will be done after we make sure release also works as expected:

- [ ] https://github.com/grafana/profiles-drilldown/issues/315

ifrost:
> Ensure that you have no stack overrides in ops / staging for your plugin

Yep, all ops instances use 1.7.0

PR #575, opened 2025-08-20
ci: Use shared action to create a changelog

ifrost:
Switching over to use shared changelog action

The formatting will change a bit but we can tweak it later. We could keep using standard-changelog though it requires adding additional yarn install before the bump-version (to actually use standard-changelog). Instead, it's probably just gonna be easier to stay consistent with other apps.


joey-grafana:
Not strong use cases really but for testing or if you didn't want to do a full release / didn't need a changelog but just seeing if the release works ;) 

I think merge away, we can adjust later if necessary.

PR #579, opened 2025-08-22
feat: switch cloud-home-app plugin proxy usage to own

MattIPv4:
### ✨ Description

Related issue(s): https://github.com/grafana/hg-cloud-home-admin/issues/997 (🔐)

`cloud-home-app` is going away in Grafana Cloud, but this plugin still has a dependency on it in code, relying on one of the plugin proxy routes it exposes, rather than using a plugin proxy route exposed by this plugin itself.

⚠️ Based on https://github.com/grafana/stack-state-service/blob/983824ea8adcb0305133179face5949cb117ffef/pkg/grafanacloud/integrations/pyroscope/pyroscope.go#L39-L45 (🔐), `.SecureJsonData.gcomApiUrl` + `.SecureJsonData.stackId` should exist; however, `.SecureJsonData.gcomApiToken` does not currently exist. Before this can be landed, the provisioning logic in Grafana Cloud for this plugin will need to be updated to generate a gcom API token that has access to the instances endpoint.

For reference, this is `cloud-home-app`'s config: https://github.com/grafana/stack-state-service/blob/983824ea8adcb0305133179face5949cb117ffef/pkg/grafanacloud/integrations/cloudhome/cloudhome.go#L17-L23 (🔐) (though I suspect this plugin could go with a stack realm token) + https://github.com/grafana/stack-state-service/blob/983824ea8adcb0305133179face5949cb117ffef/pkg/grafanacloud/integrations/cloudhome/cloudhome.go#L57-L72 (🔐) + https://github.com/grafana/auth/blob/61cd8a98818edf31bf10d1505bb014320eea7057/pkg/api/models/sets.go#L99-L104 (🔐) (though I suspect this plugin would only need stacks read in its auth set).

_This also slightly changes the hook so that the gcom API call is only made when the plugin detects it is running in Grafana Cloud, so that OSS users of the plugin don't see a random failing request, as the gcom API doesn't exist for them._

### 📖 Summary of the changes

- Added a new plugin proxy route mirroring what was defined in `cloud-home-app`
- Updated the API endpoint being called to use the new plugin proxy route
- Updated the API hook logic to only execute the fetch when running in Grafana Cloud

### 🧪 How to test?

With the correct provisioning changes in place for Grafana Cloud, confirm that the plugin running this updated logic still shows the correct Grafana Cloud onboarding with the correct settings URL for the specific stack.


MattIPv4:
Fwiw, given the super limited usage of the gcom API + returned data here, as an alternative to this API call and needing to provision a new gcom token for this plugin, the provisioning logic could instead be updated to directly inject the `orgSlug` + `hpInstanceId` into the plugin's `JSONData` (not `SecureJSONData`), which can be accessed client-side with `usePluginContext().meta.jsonData` from `@grafana/data`.

PR #597, opened 2025-09-19
fix: Update grafana/assistant

### LINKED ISSUE #596: bug: Flame Graph doesn't load in version 1.9 with Grafana 11.6
i get this error

Image
when trying to press the flame graph view in the profile drill down view
when manually downgrading the plugin it returns working
grafana version 11.6

ifrost:
fixes #596 

This will include the fix for calling a function that is not available in Grafana 11.6 (https://github.com/grafana/grafana-assistant-app/pull/1239)

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "@bufbuild/buf": "^1.50.0",
     "@bufbuild/protoc-gen-es": "^2.2.3",
     "@connectrpc/protoc-gen-connect-query": "^2.0.1",
-    "@grafana/assistant": "0.0.12",
+    "@grafana/assistant": "0.0.17",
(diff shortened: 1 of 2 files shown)


joey-grafana (approved):

LinAnt:
Is there and easy way to patch this locally? We use pyroscope quite extensively, so this hits us quite badly :)

ifrost:
> Is there and easy way to patch this locally? We use pyroscope quite extensively, so this hits us quite badly :)

@LinAnt We've released 1.10.0 we should work with 11.6.0 now

LinAnt:
Confirmed working :)

PR #626, opened 2025-10-17
feat: adaptive profiles boost button extension point

darrenjaneczek:
### ✨ Description

Adaptive profiles provides the capability of "boosting" profiles for a given service. This PR adds an extension point for an exposed component button which brings this capability to profiles drilldown.

image

image

### 📖 Summary of the changes

Expose a button on the timeseries panel when viewing a specific service. Pass that service to the button extension, so the user can execute a "boost" on it.

The extension point is expressed as a subdirectory component for the `SceneExploreServiceFlameGraph`.

**This PR only controls _WHERE_ the exposed extension will go, and what information will be passed to it. The content of the exposed extension itself will be controlled by development on the `grafana-adaptiveprofiles-app`.**

### 🧪 How to test?

To be discussed internally.

### DIFF
--- a/.github/workflows/plugins-pr.yml
+++ b/.github/workflows/plugins-pr.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
     inputs:
       branch:
-        description: Branch to publish from. Can be used to deploy PRs to dev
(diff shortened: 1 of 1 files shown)


portertech:
Perhaps the button should read "Boost Resolution"?

PR #735, opened 2025-11-05
chore(ci): Push PR branches to ops catalog

ifrost:
https://github.com/grafana/profiles-drilldown/pull/579 was tested in dev which has new endpoint available but to check what's gonna happen in cases where the version is used without the new endpoint we also need to push the branch to ops catalog


.github/workflows/plugins-pr.yml
ifrost:
```suggestion
        description: Branch to publish from. Can be used to deploy PRs to dev and ops
```

PR #744, opened 2025-11-11
fix: use local pyroscope at port 4040 when server:local

alsoba13:
### ✨ Description

I couldn't manage to connect with local pyroscope when running `yarn server:local`, I fixed it.

### 📖 Summary of the changes

Two changes:
* Fix provisioned local datasource to target the expected local pyroscope url.
* Fix load generator prefix so it sends requests to expected rideshare app

### 🧪 How to test?

1. Run pyroscope locally.
2. Then run `yarn dev` and `yarn server:local`, check how you can actually navigate to pyroscope data at localhost:3000

These steps don't work in main branch.

### DIFF
--- a/docker-compose.local.yaml
+++ b/docker-compose.local.yaml
@@ -40,7 +40,7 @@ services:
       context: ./samples/rideshare
       dockerfile: Dockerfile.load-generator
     environment:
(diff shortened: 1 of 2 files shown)


simonswine (approved):
Works for me

PR #759, opened 2025-11-27
fix: Do not force refresh ServiceName and ProfileMetric variables on time range change

ifrost:
### ✨ Description

Fixes: https://github.com/grafana/profiles-drilldown/issues/750

### 📖 Summary of the changes

After updating scenes the variable dependency is triggered when time range changes (correct because ServiceName na ProfileMetric variables are refreshed on time range changes) so instead of variable dependency config we should switch over to checking if the value of the variable actually changed.

### 🧪 How to test?

Filters should not reset when query is rerun or page is refreshed.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneMainServiceTimeseries.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneMainServiceTimeseries.tsx
@@ -5,7 +5,6 @@ import {
   SceneObjectBase,
   SceneObjectState,
   SceneQueryRunner,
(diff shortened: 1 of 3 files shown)


ifrost:
> Is there much work to get the filters added in the Asserts integration for Profiles Drilldown? I think this was the one thing we wanted to get merged first.

This fix will work for embedded mode as well.

joey-grafana:
Brilliant - I'll see if I can get a PR opened so this week to also add the filters to embedded mode

PR #784, opened 2025-12-20
fix: Sync time range sync between Diff page and the other pages

bryanhuhta:
related: https://github.com/grafana/profiles-drilldown/issues/580

# Overview

This PR changes the semantics of syncing the time range between the other explore pages in Profiles Drilldown and the diff page. In the interest of keeping changes as localized as possible, this PR *does not* address filters syncing between the diff page and the rest of the app. However, very similar sync semantics should be applied to those as well.

See a demo [here](https://github.com/user-attachments/assets/f57a77bd-f7fb-4f5a-acdb-fea96fc3f25e) for those of you who are visual learners.

**EDIT** New demo: https://github.com/user-attachments/assets/f57a77bd-f7fb-4f5a-acdb-fea96fc3f25e

# Current behavior

Historically, the diff page would maintain its own time ranges for both the baseline and comparison. This was undesirable because it caused confusion when a user moved from other places in the app to the diff page as all existing state would be lost. For example, When browsing the flame graph page with the time range set to "Last 3h", navigating to the diff page would unexpectedly reset the time range to "Last 1h".

Here is a diagram of how the time ranges were used in the old diff view:

old

# New behavior

This PR changes the syncing semantics such that the baseline panel will always reflect the global state. Changing the time range any where in the app will be reflected in the baseline panel. Likewise, changing the time range of the baseline panel will change the global time range for the entire app. The comparison panel can now set its own internal time range. Once set, navigating to other pages in the app and back will not disrupt the comparison panel's time range. The initial time range for the comparison panel will always be the global time range.

Here is a diagram of how the new time ranges are used in this PR:

new

# Other changes

While working on this PR, there were a variety of incidental changes I made as well. Some where to improve the usage experience and some where to help implement the new syncing behavior.

Change: Sync time range does not maintain a constant sync. It does a one-time sync of the time range from one panel to the other.

Change: Annotations (selecting sections of each timeline) now properly save and load from the URL. I believe at one point this was true, but somewhere along the way this feature broke. Copying and pasting a url would not respect the timeline annotations. This PR restores that feature. (It even works if you only have one timeline annotated!)

Change: The preset dropdown got a facelift. It simplifies the information it presents, it keeps it state when selecting an option (to an extent), and I removed the notion of a "custom preset". Custom presets were an idea tossed around at the beginning of this app where users could save their own preset time ranges. This idea never really gained traction, nor have customers ever asked for it so I removed it.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/SceneComparePanel.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/SceneComparePanel.tsx
@@ -15,12 +15,14 @@ import {
   SceneTimeRange,
   SceneTimeRangeLike,
   SceneTimeRangeState,
+  SceneVariableValueChangedEvent,
   VariableDependencyConfig,
 } from '@grafana/scenes';
 import { IconButton, useStyles2 } from '@grafana/ui';
 import { SceneTimePickerWithoutSync } from '@shared/components/SceneTimePickerWithoutSync/SceneTimePickerWithoutSync';
 import { getProfileMetric, ProfileMetricId } from '@shared/infrastructure/profile-metrics/getProfileMetric';
 import React from 'react';
+import { Unsubscribable } from 'rxjs';
 
 import { buildTimeRange } from '../../../../domain/buildTimeRange';
 import { FiltersVariable } from '../../../../domain/variables/FiltersVariable/FiltersVariable';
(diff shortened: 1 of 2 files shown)


simonswine:
I have managed to update some of the screenshots in this [commit](https://github.com/grafana/profiles-drilldown/commit/3c09054eee1aa817e4422d65797a988dc6c6b61e) (sorry there is a bit more mess going on), but it seems the only change is the text for when there is no selection: https://github.com/grafana/profiles-drilldown/blob/72f3a25fff364fe1362542772461604e753e83a7/src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/ui/MissingSelectionsBanner.tsx

PR #785, opened 2025-12-30
fix: Handle missing location lines from profiles

jake-kramer:
Ingested otel ebpf profiles have this structure, and were causing exceptions in the `Function details` view.


korniltsev-grafanista (approved):
I hit this couple of times.

simonswine (approved):
I have also sporadically seen profiles shaped like this.

This fixes the problem when I compile this locally

jake-kramer:
@ifrost would appreciate your review on this when you have a moment!

ifrost (approved):
Great stuff! Thanks @jake-kramer!

PR #815, opened 2026-02-11
chore: Remove husky postinstall hook

simonswine:
This is a very risky practice, I want precommit hooks to be always opt-in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
     "e2e:local:watch": "yarn e2e:local --ui",
     "format:fix": "prettier --write .",
     "generate:pyroscope-api": "npx @bufbuild/buf generate",
-    "postinstall": "husky install",
     "knip": "knip",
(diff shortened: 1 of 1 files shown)


joey-grafana (commented):
What do you mean by risky? 

This basically runs a [linter/formatter](https://github.com/grafana/profiles-drilldown/blob/main/lint-staged.config.mjs) before code is committed which keeps the code tidy by using common standards for all devs.

joey-grafana:
Technically any package can have a postinstall but I see what you're saying about our Husky postinstall and it going into git hooks.

We can remove that in favor of a CI job if you like, but can you add that workflow to this PR please? Also we will need to remove the .husky dir and it's files + the yarn package.

joey-grafana:
Fixing this in https://github.com/grafana/profiles-drilldown/pull/984

PR #822, opened 2026-02-13
Default comparison view filters to existing flamegraph/labels view filters

### LINKED ISSUE #580: [bug] Opening the diff view loses filters and time range applied in labels and flame graph views
### Bug description

- **Grafana version:**: Any
- **Plugin version:**: Any (bug existists since forever

### Steps to reproduce

1. Open a service
2. Go to Labels or Flamegraph
3. Add a filter for e.g. pod
4. Go to diff, you can see the filter got  lisot
(5.  Going back makes the filter return, as it is store in the ULR)

#### Expected behaviour

When going to the diff view the label filters and time range should be preserved.

Leaving the diff for flamegraph / labels, I am not too sure if we should take the matching filters from Left and Right of the diff view.

https://github.com/user-attachments/assets/ab148611-5b90-4a98-8ec6-9384a1a82985

bobrik:
### ✨ Description

Fixes: #580

This is probably not the most elegant way of doing this, but it does the job and it scratches a pretty big itch.

### 🧪 How to test?

1. Open labels view, select some filters.
2. Open diff view, observe filters applied to both panels.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneProfilesExplorer/SceneProfilesExplorer.tsx
@@ -345,6 +345,14 @@ export class SceneProfilesExplorer extends SceneObjectBase<SceneProfilesExplorer
       this.resetVariables(type);
     }
 
+    if (type as ExplorationType == ExplorationType.DIFF_FLAME_GRAPH && bodySceneOptions === undefined) {
+      const filters = sceneGraph.findByKeyAndType(this, 'filters', FiltersVariable).state.filters;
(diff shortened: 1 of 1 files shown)


bryanhuhta:
Just getting back to this after a short hiatus. I think with this change going in https://github.com/grafana/profiles-drilldown/pull/947, the filtering is going to need a bit more thought.

At a minimum, there needs to be some logic in place to make sure the All Services filters get copied to and removed from the flame graph and diff view filter variables.

More ideally, we handle this filter transition more akin to how the time ranges are handled when switching between flame graph and diff flame graph views.

After some discussion internally, I'm going to close this PR out, but open a new one to address the problem this PR solves while also handling more filter state book keeping for diff view.

PR #842, opened 2026-02-24
feat(adhoc): add diff view with side-by-side and computed diff modes

aleks-p:
## Summary

- Wire up the frontend to call the new `AdHocProfileService/Diff` backend endpoint (from https://github.com/grafana/pyroscope/pull/4850)
- Replace the "Comparison view" tab with a **Diff view** tab that has a toggle between **Side by side** (two independent flamegraphs) and **Diff flamegraph** (computed diff)
- Profile type selectors (one per panel) are synced across both modes — changing either one updates both views

## Test plan

- [x] Open Ad Hoc view → "Diff view" tab
- [x] Drop two pprof files (left and right), verify side-by-side flamegraphs render
- [x] Toggle to "Diff flamegraph", verify computed diff renders
- [x] Toggle back to "Side by side", verify flamegraphs are still there
- [x] Change profile type in either selector, verify both panels and diff update
- [x] Remove one file, verify both side-by-side and diff flamegraphs disappear
- [x] Re-upload a file, verify the diff recomputes automatically
- [x] Verify "Single view" tab still works independently

Video demo:

https://github.com/user-attachments/assets/1486c8eb-6cd3-41c9-a492-a501a74e0bcf

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/locales/cs-CZ/grafana-pyroscope-app.json
```diff
@@ -664,4 +664,4 @@
       "label": ""
     }
   }
-}
```
aleks-p, line 667: not sure why these files get reformatted locally, if needed I can revert them

aleks-p:
Thanks @joey-grafana!

Fundamentally this is the same functionality as the existing "Diff flame graph" view, but for manually uploaded (as opposed to continuously collected) profiles. 

There are some docs about the Diff view at https://grafana.com/docs/grafana/latest/visualizations/simplified-exploration/profiles/choose-a-view/#diff-flame-graph, though they are probably due for an update.

PR #850, opened 2026-03-04
chore(cd): remove github-draft-release

L2D2Grafana:
### ✨ Description
Related issue(s): https://github.com/grafana/logs-drilldown/pull/1785
In Logs Drilldown we set `github-draft-release: false` in our cd.yml and ran into issues with the version number zip files not being uploaded into the release. In case you run into that error as well. I've reverted the change in this PR.


joey-grafana:
Is it the same issue for PD? https://github.com/grafana/profiles-drilldown/pull/825 :)

joey-grafana (approved):

PR #934, opened 2026-04-09
feat: Profiles to Traces integration via span heatmap

simonswine:
https://github.com/user-attachments/assets/429a5acf-7ad6-489e-b522-45a5559b0c7a

## What

Adds a span profiles heatmap embedded in the flame graph view, enabling exploration from trace spans to their corresponding profiles.

When the `profilesHeatmap` feature flag is enabled, a **Span profiles** toggle button appears in the flame graph panel header. Clicking it:

1. Replaces the main timeseries with a heatmap of span profile durations (y-axis = duration bucket, x-axis = time).
2. Shows a table of top span exemplars below the heatmap, with span ID, span name, value, trace ID, and duration columns.
3. Selecting a span in the heatmap or table filters the flame graph below to that specific span profile.
4. If a Tempo datasource is configured, trace IDs are resolved from span IDs and a trace drawer can be opened inline.

image

## Feature flag

Set `GF_FEATURE_TOGGLES_ENABLE=profilesHeatmap` (also enable `profilesExemplars` for full exemplar support in the timeseries).


jake-kramer:
Great feature, looking really good.

Some UX feedback:

- A nice feature would be to auto select the Tempo datasource if there's only one present.
- I would expect clicking on a value in the `Trace ID` column in the `Top span exemplars` table to open the Trace.
- Moving to different drilldown tabs (like `Labels` or `Diff flame graph`) resets state in `Flame graph` tab such that span profiles is disabled and Tempo datasource deselected.
- Selecting an exemplar will highlight the trace in the `Top span exemplars` table, but won't scroll the table down to that line if there's enough entries to make the table scrollable.
- Selecting an exemplar add the span id to the flame graph filter (makes sense) but I wonder if changing the time window selector should clear/reset the filter? I had a exemplar selected, then wanted to zoom with a dragging window in to some separate part of the timeline but the flame graph didn't update (had old span id still selected).
- Probably out of scope for this PR but both the exemplars and span profiles feature buttons feel super easy to miss. Wondering if there's a way to make them more prominent (and perhaps with a little `i` info hover to briefly explain the feature to users)
- Out of scope for this PR, but would be very cool to integrate some sort of way to select two traces/spans to jump into the diff flame graph view with.

L2D2Grafana:
👋 Hello any tips for running this locally or with Graft? I tried [dev](https://dev.grafana-dev.net/a/grafana-pyroscope-app/explore?searchText=&panelType=time-series&layout=grid&hideNoData=off&explorationType=flame-graph&var-serviceName=alertmanager%2Falertmanager&var-profileMetricId=process_cpu:cpu:nanoseconds:cpu:nanoseconds&var-groupBy=&var-profileIdSelector=&var-spanSelector=undefined&var-dataSource=af2bv2jaln0n4e&var-filters=&var-filtersBaseline=&var-filtersComparison=&diffFrom=&diffTo=&diffFrom-2=&diffTo-2=&comparisonFrom=&comparisonTo=&from=now-2d&to=now&timezone=browser&maxNodes=16385) with graft and locally with the default setup but can't see the span profiles yet, ty!

simonswine:
@L2D2Grafana thanks for taking the time, in order to get this working locally:

I am using these changes: https://github.com/grafana/profiles-drilldown/commit/aeaac1a018f5d23bcac735c50b117845b7df1c4a

Then in .env I am using the fire-dev-001 cell.

And configure tempo like that:

```
REMOTE_TRACE_BACKEND_URL=https://tempo-dev-01-dev-us-central-0.grafana-dev.net/tempo
REMOTE_TRACE_BASIC_AUTH_USER=3
REMOTE_TRACE_BASIC_AUTH_PASSWORD=glc_dev_xxx # same as profiels for me
```

And then running `pnpm dev --no-watch && pnpm server:remote` sets it all up.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneExploreServiceFlameGraph.tsx
```diff
@@ -102,21 +350,84 @@ export class SceneExploreServiceFlameGraph extends SceneObjectBase<SceneExploreS
...
+          <favAction.Component model={favAction} />
+          <SpanExemplarToggleButton model={spanToggle} />
+        </>
+      }
+    >
+      <spanHeatmap.Component model={spanHeatmap} />
+    </Panel>
+  );
+}
+
```
L2D2Grafana: This could be nice as a separate component SpanHeatmapPanel

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SpanSelectorLabel.tsx
```diff
@@ -22,27 +25,97 @@ export function SpanSelectorLabel(props: Props) {
...
+        name="="
+        onClick={noOp}
+        tabIndex={0}
+      />
+
+      <Tag
+        aria-label={t('flame-graph.span-selector.filter-value', 'Filter value')}
+        name={spanSelector}
+        className={styles.noInteraction}
+        onClick={noOp}
```
L2D2Grafana: Do these tags do anything? Where is this in the ui?

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SpanSelectorLabel.tsx
```diff
@@ -22,27 +25,97 @@ export function SpanSelectorLabel(props: Props) {
...
+      color: ${theme.colors.text.maxContrast};
+    }
+
+    & > :first-child {
+      background-color: ${activeBackgroundColor};
+      color: ${activeTextColor};
+      border-radius: 0;
+
+      &:hover {
+        cursor: not-allowed !important;
```
L2D2Grafana: Usually you want a cursor on hover.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneExploreServiceFlameGraph.tsx
```diff
@@ -46,16 +97,213 @@ export class SceneExploreServiceFlameGraph extends SceneObjectBase<SceneExploreS
...
+        this.closeSpanHeatmapMode();
+      }
+    });
+
+    // When spanSelector changes from outside (e.g. the "×" button on SpanSelectorLabel),
+    // sync the selection into the visible heatmap.
+    const spanSelectorSub = sceneGraph
+      .findByKeyAndType(this, 'spanSelector', SpanSelectorVariable)
+      .subscribeToState((newState, prevState) => {
+        if (newState.value === prevState.value || !this.state.spanHeatmap || this.syncingHeatmapSelection) {
```
L2D2Grafana, line 150: This syncingHeatmapSelection might keep stale span filters when changing the time window or tabs.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExploreServiceHeatmap.tsx
```diff
@@ -0,0 +1,319 @@
...
+  return {
+    dataSourceUid,
+    profileTypeId,
+    request,
+    signature: JSON.stringify({ dataSourceUid, request }),
+  };
+}
+
+export function buildSpanHeatmapState(response: SelectHeatmapResponse, profileTypeId: string) {
+  const { unit } = getProfileMetric(profileTypeId as any);
```
L2D2Grafana: We try not to use `as any` and use typescript types.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/infrastructure/buildHeatmapDataFrames.ts
```diff
@@ -0,0 +1,301 @@
...
+ * Builds an annotation DataFrame for heatmap exemplar markers, matching the format
+ * produced by exemplar/exemplar.go (PR #120995).
+ *
+ * Includes all label fields (series labels merged with per-exemplar labels) so the
+ * heatmap tooltip shows the same context as the exemplar popover in the timeseries view.
+ *
+ * The heatmap panel picks up frames with `meta.dataTopic === 'annotations'` and
+ * `frame.name === 'exemplar'` to render as diamond markers.
+ */
+// eslint-disable-next-line sonarjs/cognitive-complexity
```
L2D2Grafana, line 221: Wow never seen this rule before. It does seem pretty intense to do this many for loops on the UI side. It make me wonder about formatting it in the backend or running it in a webworker, if the ui slows down.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/infrastructure/HeatmapApiClient.ts
```diff
@@ -0,0 +1,41 @@
...
+  end: number;
+  /** Query resolution step width in seconds */
+  step: number;
+  groupBy: string[];
+  queryType: HeatmapQueryType;
+  exemplarType: ExemplarType;
+  limit?: number;
+}
+
+export class HeatmapApiClient extends DataSourceProxyClient {
```
L2D2Grafana: Do we need a class for an api call? I'd think we could just call an async function

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
```diff
@@ -0,0 +1,616 @@
...
+    const requestId = this.traceLookupRequestId;
+    const requestTempoDataSourceUid = tempoDataSourceUid;
+
+    const conditions = spanIds.map((id) => `span:id="${id}"`).join(' || ');
+    const query = `{${conditions}}`;
+
+    this.setState({ loadingSpanIds: [...this.state.loadingSpanIds, ...spanIds] });
+
+    // eslint-disable-next-line sonarjs/cognitive-complexity
+    (async () => {
```
L2D2Grafana, line 166: This seems like a nice chunk of code to abstract.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
```diff
@@ -0,0 +1,616 @@
...
+  loadingSpanIds: string[];
+  tempoDatasources: TempoDatasource[];
+  tempoDataSourceUid?: string;
+}
+
+function formatTimestamp(ms: number): string {
+  return new Date(ms).toLocaleTimeString();
+}
+
+function CopyableId({ value, display }: { value: string; display: string }) {
```
L2D2Grafana: This could be abstracted.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
```diff
@@ -0,0 +1,616 @@
...
+
+    return (
+      <div className={styles.container}>
+        <div className={styles.header}>
+          <span className={styles.title}>{t('heatmap.exemplar-table.title', 'Top span exemplars')}</span>
+          <div className={styles.picker}>
+            <FormField
+              label={t('heatmap.exemplar-table.tempo-datasource', 'Tempo datasource')}
+              className={styles.pickerField}
+            >
```
L2D2Grafana, line 438: I couldn't get this working yet even with 
```
REMOTE_TRACE_BACKEND_URL
```

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
```diff
@@ -0,0 +1,616 @@
...
+          </div>
+        </div>
+
+        {rows.length === 0 ? (
+          <div className={styles.empty}>
+            {t('heatmap.exemplar-table.empty', 'No span exemplars in the selected time range.')}
+          </div>
+        ) : (
+          <div className={styles.tableWrapper}>
+            <table className={styles.table}>
```
L2D2Grafana: It is hard to see the highlighted exemplar if it's below the fold. Since it's all being filtered anyway make instead of returning the table we just show that one row highlighted.  Also we could try and use the table in grafana/ui here.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExploreServiceHeatmap.tsx
```diff
@@ -0,0 +1,319 @@
...
+      this.primedResponse = undefined;
+      this.setState({ isLoading: false, ...heatmapState });
+      return;
+    }
+
+    this.setState({ isLoading: true });
+
+    try {
+      const client = new HeatmapApiClient({ dataSourceUid: spanHeatmapQuery.dataSourceUid });
+      const response = await client.selectHeatmap(spanHeatmapQuery.request);
```
L2D2Grafana: I like the abstraction but it could just be a function.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneHeatmapMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { quoteLabelName } from '@shared/components/QueryBuilder/domain/helpers/quoteLabelName';
+import React from 'react';
+
+import { FiltersVariable } from '../../domain/variables/FiltersVariable/FiltersVariable';
+import { getExploreUrl } from '../../helpers/getExploreUrl';
+import { TimeSeriesQuery } from '../../infrastructure/timeseries/buildTimeSeriesQueryRunner';
+
+export class SceneHeatmapMenu extends SceneObjectBase<SceneObjectState> {
+  onClickExplore() {
+    reportInteraction('g_pyroscope_app_open_in_explore_clicked');
```
L2D2Grafana, line 19: Nice! We should add reportInteractions for users clicking on the span profiles and exemplars to see how often people use this.

L2D2Grafana (commented):
🥳 This PR is awesome! I know it's super hard adding stuff and interactions to the graphs 👏 . Feel free to disregard any of my comments as I do not know all the rules of this repo. 

A few miscellaneous ux things:

- Moving or duplicating the spanid filter to the top query editor experience might help make a few interactions easier. Users could see the id in the graph and select that id from the filter at the top. It's a bit tricky to actually click an exemplar then click again to select the exemplar and filter on it.

Image

- Scrolling the top exemplar table and the full page compete with each other. Using the grafana table with pagination could be a solution.

src/pages/ProfilesExplorerView/domain/actions/SpanExemplarToggleAction.tsx
```diff
@@ -0,0 +1,59 @@
...
+            )
+          : showSpanExemplars
+          ? t('actions.span-exemplar-toggle.hide', 'Hide span profiles')
+          : t('actions.span-exemplar-toggle.show', 'Show span profiles')
+      }
+      tooltipPlacement="top"
+      onClick={model.onClick}
+    >
+      <Trans i18nKey="actions.span-exemplar-toggle.label">Span profiles</Trans>
+    </Button>
```
bryanhuhta: I think the tooltip is unnecessary here. The tooltip text should be the button text.

src/pages/ProfilesExplorerView/domain/actions/SpanExemplarToggleAction.tsx
```diff
@@ -0,0 +1,59 @@
...
+  };
+}
+
+export function SpanExemplarToggleButton({ model }: { model: SpanExemplarToggleAction }) {
+  const { showSpanExemplars, hasSpanData } = model.useState();
+  const disabled = hasSpanData === false;
+
+  return (
+    <Button
+      icon={showSpanExemplars ? 'eye' : 'eye-slash'}
```
bryanhuhta: I think this should be reversed. We use the eye icon if there are no span exemplars shown and the eye-slash icon if there are exemplars shown. This will help communicate the action the button performs.

Default view, no span exemplars shown:
Image

With span exemplars shown:
Image

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneExploreServiceFlameGraph.tsx
```diff
@@ -102,21 +350,84 @@ export class SceneExploreServiceFlameGraph extends SceneObjectBase<SceneExploreS
...
+      isLoading={isLoading}
+      menu={() => <menu.Component model={menu} />}
+      headerActions={
+        <>
+          <ResolutionBoostExtensionPoint scene={model} />
+          <selectAction.Component model={selectAction} />
+          <favAction.Component model={favAction} />
+          <SpanExemplarToggleButton model={spanToggle} />
+        </>
+      }
```
bryanhuhta: I can see that we are quickly going to have header actions being very messy and unorganized. I think these worked fine for when there was just the "favorite" and occasional "Labels" or "Flame graph" link, but now its getting very full.

Here is what it looks like in ops:
Image

Adding another "Show span profiles" button puts a lot of actions in this one spot.

Image

I personally think we should drop the "Labels" and Favorites icon button. They're styled oddly compared to the other header actions and there are other more obvious ways to achieve the same results (or in the case of Favorites, it was never a popular feature to begin with).

src/locales/en-US/grafana-pyroscope-app.json
```diff
@@ -328,6 +344,34 @@
...
+      "col-timestamp": "Timestamp",
+      "col-trace-id": "Trace ID",
+      "copied": "Copied!",
+      "copy-to-clipboard": "Copy to clipboard",
+      "empty": "No span exemplars in the selected time range.",
+      "open-flamegraph": "Open flame graph",
+      "open-trace": "Open trace",
+      "select-tempo": "Select Tempo datasource",
+      "tempo-datasource": "Tempo datasource",
+      "tempo-placeholder": "Select Tempo datasource…",
```
bryanhuhta: I believe its "data source" (2 words) and not "datasource" (1 word).

bryanhuhta (commented):
Compiled most of my thoughts in the video below. I like where this is at, in general. There's a few rough edges from a usability perspective, but I think it's pretty close to something we can ship.

A thought I had after recording the video: I'm not sure "Show/Hide span profiles" is the right text for what is essentially switching the visualization to/from a heatmap. Perhaps we want to use something like "Heatmap/Time series" instead.

https://github.com/user-attachments/assets/f2baf6af-0fe2-437a-a018-0f25bb760b9e

L2D2Grafana:
I can't seem to get the traces working locally even with the extra provisioning.
Screenshot 2026-07-16 at 8 53 49 AM One thing I've been curious about is adding profiles to our fake data generator https://github.com/grafana/logs-drilldown/pull/1748. I'm not sure exactly what needs to be added or how we would create anomalies or exemplars. If you have any advice on this side quest, lmk 🙏

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneExploreServiceFlameGraph.tsx
simonswine: done

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SpanSelectorLabel.tsx
simonswine: Was still an AI slop artefact removed

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SpanSelectorLabel.tsx
simonswine: removed

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneExploreServiceFlameGraph.tsx
simonswine: I think this is clearing now

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExploreServiceHeatmap.tsx
simonswine: Done, there are still others to refactor

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/infrastructure/buildHeatmapDataFrames.ts
simonswine: Agree this should be a backend task, currently I am not fully sure if that is even correct, thats why I am doing it here before doing a grafana/grafana PR afterwards.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/infrastructure/HeatmapApiClient.ts
simonswine: Replaced it

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
simonswine: This is now in `infrastructure/lookupTempoTraceInfo.ts`

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
simonswine: Extracted this into `components/CopyableId.tsx`.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
simonswine: TODO: Will provide a separate PR with the changes for making REMOTE_TRACE_BACKEND_URL work

In the meantime I have deployed it here: https://pyro.grafana-ops.net/

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
simonswine: Implemented using the Grafana table

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExploreServiceHeatmap.tsx
simonswine: done

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneHeatmapMenu.tsx
simonswine: Added

- `g_pyroscope_app_span_exemplar_selected`
- `g_pyroscope_app_open_in_explore_clicked`

src/pages/ProfilesExplorerView/domain/actions/SpanExemplarToggleAction.tsx
simonswine: Now using tooltip only for the disabled no-data state.

src/pages/ProfilesExplorerView/domain/actions/SpanExemplarToggleAction.tsx
simonswine: Did it quite a bit differently, let me know what you think:

image

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneExploreServiceFlameGraph.tsx
simonswine: Remove labels, added favourites to the menu, potentially exemplars should probably move the menu too.

image

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneTracePanel.tsx
```diff
@@ -0,0 +1,160 @@
...
+    };
+  }
+
+  private buildPanel(spanId?: string): SceneObject {
+    const panel = PanelBuilders.traces().setHoverHeader(true);
+    if (spanId) {
+      panel.setOption('focusedSpanId' as any, spanId as any);
+    }
+    return panel.build();
+  }
```
simonswine, line 96: TODO: Need to still get rid of those any casts

src/locales/en-US/grafana-pyroscope-app.json
simonswine: Done

simonswine:
Thanks for all your feedback. I hope I have not missed too much.

I removed the orange diamond because it is not maintainable here; restoring it would require upstream work. When you select an exemplar, it turns orange and the other exemplars are no longer highlighted.

@L2D2Grafana I kept the span selector location out of scope for this PR because it requires related work on the profile ID selector, which is currently presented the same way. I created a follow-up issue: https://github.com/grafana/profiles-drilldown/issues/1053

@bryanhuhta Thanks for the video walkthrough.
Regarding hover popups: my understanding is that we cannot reuse the time-series implementation for the heatmap. Supporting an expanded action area or a shared popup for both a bucket and its exemplars would require changes in grafana/grafana, so I think we should address this separately.

@jake-kramer The Tempo data source should now auto-populate and persist: it prefers the source linked through the traces-to-profiles configuration, then falls back to the only available Tempo source. I also hope the radio button makes the visualization mode more discoverable.

In the meantime I have deployed it here: https://pyro.grafana-ops.net/ so it's easier for folks to try

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
L2D2Grafana: I gave it a try here https://github.com/grafana/profiles-drilldown/pull/1054. I was also wondering about the LLM App. If most things have been migrated to the assistant now, perhaps we could remove it 🤔

src/pages/ProfilesExplorerView/domain/actions/SpanExemplarToggleAction.tsx
L2D2Grafana: ❤️ Span Heatmap

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
```diff
@@ -0,0 +1,545 @@
...
+            {t('heatmap.exemplar-table.trace-lookup-failed', 'Unable to load trace details from Tempo.')}
+          </div>
+        )}
+
+        {displayedRows.length === 0 ? (
+          <div className={styles.empty}>
+            {t('heatmap.exemplar-table.empty', 'No span exemplars in the selected time range.')}
+          </div>
+        ) : (
+          <InteractiveTable
```
L2D2Grafana, line 464: Very cool! 

If you select an exemplar at the bottom of the graph the table can disappear now. It's weird because if you select from the table and filter instead of the heatmap it seems to work.
Image 

I'd also make the height a bit more so that the scrollbar goes away and the pagination is always visible
Image

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
```diff
@@ -0,0 +1,545 @@
...
+                  tooltip={t('heatmap.exemplar-table.open-flamegraph', 'Open flame graph')}
+                  size="sm"
+                  onClick={() => model.openFlameGraph(row.profileId, row.spanId, row.timestamp)}
+                />
+                {row.spanId && traceInfo && (
+                  <IconButton
+                    name="compass"
+                    tooltip={t('heatmap.exemplar-table.open-trace', 'Open trace')}
+                    size="sm"
+                    onClick={() => openTrace(row, 'action')}
```
L2D2Grafana, line 409: I wonder if this could be a different icon @joey-grafana is there a trace view icon? In logs we use the button with the text traces and a link icon.

L2D2Grafana (commented):
Thanks for the updates. There are a number of minor things we could do, but I think if we can fix the table not showing results for selected exemplars lower down on the graph. I'd be fine merging this and continuing on with smaller PRs. I got the trace view working and that looks good 👍 although the icon to open it is debatable.

src/pages/ProfilesExplorerView/components/SceneExploreServiceHeatmap/SceneExemplarTable.tsx
simonswine: I have previously limited the exemplars rows that are going into the table to 50. This has caused this behaviour when you would select an exemplar that didn't make the top 50.

I have removed that limit, it will likely make the tempo query a bit slower, but I think we can probably review this to only include tempo queries currently visible going forward.

Removed top 50 in 0e55d5966cab72e4f2421fc174b4bf0a670fc5de

L2D2Grafana (dismissed):
Thnx for all the updates! Since it's behind feature toggles `profilesHeatmap,profilesExemplars` etc I think it's fine to merge and continue on. Also if there is scale issues users should be able to bail and go back to the Timeseries view🫡

L2D2Grafana:
Not sure what I did to my traces, but when it doesn't work this is the error mesage
Screenshot 2026-07-21 at 2 19 50 PM

simonswine:
@L2D2Grafana just pushed an update to fix a conflict.

> Not sure what I did to my traces, but when it doesn't work this is the error mesage

I have seen this a lot more too since, we query a lot more span IDs the queries time out way more often. I want to address this separately in a follow up, by batching them up into the row you can actually see.

L2D2Grafana (approved):

PR #948, opened 2026-04-20
fix(deps): patch CVEs via yarn resolutions

nicwestvold:
### ✨ Description

Fixes four dependency vulnerabilities flagged via `yarn npm audit`.

| Advisory | Severity | Package | Installed | Fixed |
|---|---|---|---|---|
| CVE-2026-41242 | Critical | `protobufjs` | 7.3.2 | 7.5.5 |
| CVE-2026-41242 | Critical | `protobufjs` | 8.0.0 | 8.0.1 |
| CVE-2026-5758 | Moderate | `protocol-buffers-schema` | 3.6.0 | 3.6.1 |
| GHSA-v3rj-xjv7-4jmq | Moderate | `smol-toml` | 1.3.1 | 1.6.1 |

### 📖 Summary of the changes

All four packages are transitive dependencies, so the fixes are applied via `resolutions` in `package.json` - matching the existing pattern for other pinned security fixes (e.g. `serialize-javascript`, `dompurify`).

- `protobufjs` lives at two majors in the tree:
  - `7.3.2` is pulled in by `@grafana/faro-core@1.19.0 → @opentelemetry/otlp-transformer@0.202.0` (range `^7.3.0`) → scoped resolution `"protobufjs@npm:^7.3.0": "7.5.5"` keeps it on v7.
  - `8.0.0` is pinned exactly by `@grafana/faro-core@2.2.4 → @opentelemetry/otlp-transformer@0.212.0` → scoped resolution `"protobufjs@npm:8.0.0": "8.0.1"` bumps just that copy.
- `protocol-buffers-schema` (via `ol → pbf → resolve-protobuf-schema`) and `smol-toml` (via `knip`, dev-only) are safe same-major bumps, so they use global resolutions.

All four bumps stay within the same major version, so breaking-change risk is minimal.

### 🧪 How to test?

1. `yarn install`
2. `yarn npm audit --all --recursive` — only deprecation notices should remain; no CVEs.
3. `yarn typecheck && yarn test:ci && yarn build` — confirm the resolution changes don't regress the build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,11 @@
     "@tootallnate/once": "3.0.1",
     "diff": "4.0.4",
     "flatted": "3.4.2",
-    "serialize-javascript": "7.0.5"
+    "serialize-javascript": "7.0.5",
+    "protobufjs@npm:^7.3.0": "7.5.5",
+    "protobufjs@npm:8.0.0": "8.0.1",
+    "protocol-buffers-schema": "3.6.1",
+    "smol-toml": "1.6.1"
   },
   "dependencies": {
     "@bufbuild/protobuf": "^2.2.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -12578,9 +12578,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"protobufjs@npm:8.0.0":
-  version: 8.0.0
-  resolution: "protobufjs@npm:8.0.0"
+"protobufjs@npm:7.5.5":
+  version: 7.5.5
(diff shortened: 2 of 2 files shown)


joey-grafana (approved):
LGTM 👍 Eventually we will need to clean up these resolutions perhaps by upgrading the actual packages that they are used in. 

I suspect these packages are used in Core Grafana packages like data, runtime etc? 

If they are used in in data, runtime etc then this PR is fine. If they are not, can we upgrade the actual package(s) that use these instead of using resolutions?

PR #957, opened 2026-04-30
Chore: Update @grafana/flamegraph to 13.1.0-25141410377

leeoniya:
this updates flamegraph package to the most recent version, which uses `d3-scale` directly rather than pulling the entire `d3` from grafana core into every loaded dashboard.


shelldandy (approved):
Huge savings!!

PR #975, opened 2026-05-13
fix: Download json

joey-grafana:
### ✨ Description

Exporting a flame graph as JSON built a data: URL with encodeURIComponent(JSON.stringify(profile)). For large profiles that approach:

- Very high memory use (stringify + full escaped copy),
- Browser limits on data: URLs and heavy work on the main thread, which showed up as the UI locking up or the tab crashing (often more noticeable with big production profiles, e.g. as a Viewer).

Solution

- Serialize the profile with JSON.stringify, wrap it in a Blob (application/json;charset=utf-8), and call file-saver’s saveAs(blob, filename) so the download uses a normal blob URL instead of a giant inline data: string.
- Centralize that in saveProfileJsonToFile (src/shared/domain/saveProfileJsonToFile.ts) and use it from: SceneExportMenu (Profiles explorer export menu), useExportMenu (shared FlameGraph export menu).
- SceneExportMenu: wrap in try/catch and surface failures with displayError (same pattern as other export errors).

### 🧪 How to test?

Go to flame graph or diff flame graph tabs and try to export flame graph as json - it should work.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25,7 +25,6 @@ overrides:
   uuid: '>=14.0.0'
 
 importers:
-
   .:
     dependencies:
       '@bufbuild/protobuf':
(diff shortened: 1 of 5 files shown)


matyax (dismissed):

joey-grafana:
Currently fixing test, sorry for dismissing review!

matyax (approved):

PR #979, opened 2026-05-21
fix(Recording Rules): Stop dropping profile type, service name, and function name on submit

alsoba13:
## Evidence / reproduce bug
1. Create a recording rule, from a service flame graph, with a filter on it:
image
2. Message sent to backend has no service_name, no function_name, and wrong __profile_type__:
Screenshot 2026-05-21 at 11 35 19
3. The listed new rule shows this bug as well:
Screenshot 2026-05-21 at 11 35 34

## Summary

- The create-recording-rule wizard was silently dropping `profileType`, `serviceName`, and (when prefilled from the flame-graph context menu) `functionName`, sending `{ __profile_type__="undefined" }` to the backend and omitting the `service_name=` matcher and `stacktraceFilter` entirely.
- Root cause is a react-hook-form behavior change introduced in **7.60.0**: with `shouldUnregister: true`, field values supplied only through `values:` (never typed by the user) are no longer included in submitted data. The pnpm migration (#949) bumped react-hook-form from `7.54.2` to `7.72.1`, crossing that boundary.
- Bisect: 7.55.0–7.59.0 ✅, 7.60.0–7.75.0 ❌.

## Fix

One-line change: drop `shouldUnregister: true` from the modal's `useForm` config. The form has no use case for it (no dynamically appearing fields, no need to wipe state between opens — the modal unmounts entirely).

## Test plan

- [x] Added unit test that mounts the modal with a prefilled `functionName`, submits, and asserts the payload to `actions.save()` contains all five fields. Verified it fails when `shouldUnregister: true` is restored.
- [x] `pnpm exec tsc --noEmit` clean.
- [x] `pnpm exec eslint` clean (one pre-existing `MultiSelect` deprecation warning, unrelated).
- [ ] Manual smoke: open Create recording rule wizard, fill it in, confirm the request body contains the correct `__profile_type__`, `service_name=` matcher, and `stacktraceFilter`.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneCreateMetricModal/SceneCreateRecordingRuleModal.test.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneCreateMetricModal/SceneCreateRecordingRuleModal.test.tsx
@@ -1,0 +1,92 @@
+import { sceneGraph } from '@grafana/scenes';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import React from 'react';
+
+import { SceneCreateRecordingRuleModal } from './SceneCreateRecordingRuleModal';
+
+jest.mock('react-inlinesvg', () => {
+  const React = require('react');
+  return { __esModule: true, default: () => React.createElement('svg', {}) };
+});
+
+const mockSave = jest.fn();
+jest.mock('./domain/useCreateRecordingRule', () => ({
+  useCreateRecordingRule: () => ({ actions: { save: mockSave } }),
+}));
+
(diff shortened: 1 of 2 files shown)


L2D2Grafana (approved):
LGTM

This was tricky to debug. I used graft in dev to see the prom recording rules
Don't click on the top total bar in the flame graph it always works! Click on the 2 or 3 bars below it. 
Image
profiles_record_stuff3 graft had the changes and all the undefined were defined.
Screenshot 2026-05-21 at 8 55 32 AM

PR #984, opened 2026-05-26
chore: Harden package use and workspace environment

joey-grafana:
### ✨ Description

- Hardens package use and workspace environment. 
- Upgrade pnpm
- Creates pnpm-workspace file
- Follows pnpm best practices + more for hardening of environment
- Removes/upgrades some packages
- Replaces others with non-package equivalents
- Moves package overrides to pnpm-workspace file as this is required in v11+
- Adds --frozen-lockfile --ignore-scripts
- Sets minimumReleaseAge, trustPolicy, blockExoticSubdeps etc
- Removes husky postinstall hook (fixes https://github.com/grafana/profiles-drilldown/pull/815)


src/shared/domain/randomId.ts
```diff
@@ -0,0 +1,12 @@
...
+/** Short random string for React keys and ephemeral client-side IDs. */
+export function randomId(size: number): string {
+  const bytes = new Uint8Array(size);
+  crypto.getRandomValues(bytes);
+  let id = '';
+  for (let i = 0; i < size; i++) {
+    id += ID_ALPHABET[bytes[i]! % ID_ALPHABET.length]!;
+  }
+  return id;
+}
```
L2D2Grafana: Should we delete this in favor of crypto.randomUUID();

.dockerignore
```diff
@@ -1,6 +1,5 @@
 .git
 .github
-.husky
```
L2D2Grafana, line 3: Did you want to delete husky? We use it in logs drilldown and find it quite nice in finding errors before making PRs.

pnpm-workspace.yaml
```diff
@@ -1,10 +1,50 @@
...
-  - esbuild
-  - protobufjs
-  - unrs-resolver
+# https://pnpm.io/supply-chain-security
+# Same as passing --frozen-lockfile on every pnpm install (CI uses this via plugin-ci-workflows too).
+frozenLockfile: true
+# Disable package lifecycle scripts (postinstall, etc.) on install.
+ignoreScripts: true
+# One week in minutes (7 × 24 × 60).
+minimumReleaseAge: 10080
```
L2D2Grafana: I think the default grafana renovate is 3 days, but 7 days is probably fine too. https://github.com/grafana/grafana-renovate-config/blob/5617b5a4e40447fd2a48cf4a102cbeb92a5c0b96/presets/npm.json#L7

webpack.config.ts
```diff
@@ -47,6 +47,7 @@ const config = async (env): Promise<Configuration> => {
     resolve: {
       extensions: ['.ts', '.tsx', '.js', '.json'],
       alias: {
+        src: path.resolve(__dirname, './src'),
```
L2D2Grafana, line 50: Intereseting I thought this was already supported in the .config/webpack/webpack.config.ts  in the resolve modules section

L2D2Grafana (dismissed):
LGTM 👍 

Image

I'm unconvinced about radomId.ts  and removing husky.

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneComparePanel/ui/colors.ts
```diff
@@ -1,11 +1,9 @@
-import Color from 'color';
-
```
L2D2Grafana, line 2: Nice!

src/shared/domain/randomId.ts
joey-grafana: Just that crypto.randomUUID() is always 36 chars whereas randomId has size param. You're right though, don't need size param!

pnpm-workspace.yaml
joey-grafana: I thought 7 would be safer as there was discussion around 3 not being enough. I changed it to 5.

webpack.config.ts
joey-grafana: Good catch — resolve.modules already includes src, but that only helps bare imports (constants, pages/...), not src/... imports. The src alias is for the explicit src/ prefix used in a bunch of files. They’re complementary, not duplicates. @ shared / @ img are separate path aliases that aren’t covered by modules either.

.dockerignore
joey-grafana: It's not running anymore with postinstall here as we have ignoreScripts: true.

It's basically a lint script though I think - which we can still run manually.

bohandley (approved):
LGTM!

PR #1019, opened 2026-06-16
fix: replace crypto.randomUUID with fallback for non-secure contexts

### LINKED ISSUE #1017: [bug] crypto.randomUUID is not a function
### Bug description

- **Grafana version:**Grafana v13.0.1+security-01 (9bbe672d)
- **Plugin version:**2.0.7

### Steps to reproduce

1. Go to this URL: grafana in *HTTP* protocol
2. open drilldown/profiles

#### Actual behaviour

Image

#### Expected behaviour

work

#### See also

https://stackoverflow.com/questions/77980939/typeerror-crypto-randomuuid-is-not-a-function

`window.crypto.randomUUID` is only available under HTTPS context

okhowang:
### ✨ Description

Fixes #1017

### 📖 Summary of the changes

crypto.randomUUID() is only available in secure contexts (HTTPS). When Grafana is accessed over HTTP, this throws a TypeError. Replace all direct calls with a shared generateUUID() utility that falls back to crypto.getRandomValues() in non-secure contexts.

### 🧪 How to test?

run in HTTP grafana instance


src/shared/domain/generateUUID.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Generate a UUID v4.
+ *
+ * Uses `crypto.randomUUID()` when available (secure contexts / HTTPS).
+ * Falls back to a `crypto.getRandomValues()`-based implementation for non-secure contexts
+ * (e.g. Grafana accessed over HTTP) where `crypto.randomUUID` is not defined.
+ */
+export function generateUUID(): string {
```
joey-grafana, line 8: Can we add a test for this

src/shared/domain/generateUUID.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') {
+    return crypto.randomUUID();
+  }
+
+  // Fallback: RFC 4122 version 4 UUID using crypto.getRandomValues()
+  // (available in non-secure contexts, unlike crypto.randomUUID)
+  const bytes = new Uint8Array(16);
+  crypto.getRandomValues(bytes);
```
joey-grafana: We should also guard against crypto && crypto.getRandomValues being available

src/shared/domain/generateUUID.ts
okhowang: test added

src/shared/domain/generateUUID.ts
okhowang: done

joey-grafana:
hey merge with main should fix failing tests

okhowang:
> hey merge with main should fix failing tests

this PR shouldn't effect e2e test, it maybe some case is unstable?

joey-grafana:
Please don't force push, but simply merge with main to get latest changes which will fix the tests as they're fixed on main

joey-grafana (approved):

joey-grafana:
Last fix: Commits must have verified signatures :)

okhowang:
commit signed

PR #796, opened 2026-01-15
feat: inspect a single exemplar's flamegraph

alsoba13:
### ✨ Description
This PR addresses single profile filtering in flame graph views.
This PR needs a grafana with these changes: https://github.com/grafana/grafana/pull/116325

### 📖 Summary of the changes

* You can select an exemplar from the timeseries and choose to "View profile". If you are not in the flamegraph view, you will be redirected there.
* Then, flamegraph query will use profileIdSelector to only return that profile's flame graph.
* The filter is visible in the top of the flame graph panel and you can remove it anytime.

Example video:

https://github.com/user-attachments/assets/ce2c3a55-d17b-4b54-b772-48eb7dc5da46

### 🧪 How to test?
* Run grafana branch: https://github.com/grafana/grafana/pull/116325
* Run this profile drill down branch.
* Run pyroscope (no changes there, main or a version >= [v1.17.0](https://github.com/grafana/pyroscope/releases/tag/v1.17.0) should work)

PR #1042, opened 2026-07-01
fix: include profile ID selector in function details query

alsoba13:
### ✨ Description

Fixes: https://github.com/grafana/pyroscope-squad/issues/831
Related issue(s): 

When inspecting a single exemplar's flame graph (the profile ID selector added in #796), the flame graph query is scoped to that single profile via `profileIdSelector`. However, the Function Details / GitHub integration panel fetches its own data through `SelectMergeProfile` (`PprofApiClient.selectMergeProfileJson`) and did **not** pass the profile ID selector.

As a result, while the flame graph was scoped to the selected profile, the Function Details panel queried across **all** profiles matching the label selector, so users saw unexpected profile values in the panel.

This threads the `profileIdSelector` scene variable from `SceneFunctionDetailsPanel` → `useFetchFunctionsDetails` → `PprofApiClient.selectMergeProfileJson`, which now includes it in the `SelectMergeProfile` request (as an array, matching how the Grafana Pyroscope datasource sends the same field). It's added to the react-query key so the panel refetches when the selection changes, and it is only included in the request body when set (so non-flame-graph views are unaffected).

### 🧪 How to test?

1. Open a service flame graph that has exemplars.
2. Click an exemplar to inspect a single profile (adds the profile ID selector — you'll see the "profile id selector" label on the flame graph).
3. Open the Function Details panel for a function in the flame graph.
4. Before this change, the values shown were computed across all matching profiles; after this change, they match the single selected profile.


marcsanmi (approved):

L2D2Grafana (approved):
Tested in dev with graft, seems like the id is being sent.

Image

Adding a unit test or e2e would be great.

PR #1054, opened 2026-07-17
chore(remote): add remote provisioning for the llm app and traces

L2D2Grafana:
## Summary 📝

- Provisions the `grafana-llm-app` and a remote Tempo datasource for the `server:remote` dev environment, wiring span-profiles-to-traces so the span exemplar table's "Tempo datasource" picker works against remote data.
- Adds `GF_PLUGINS_PREINSTALL_DISABLED=true` across all docker-compose files so Grafana 12.3+ no longer overwrites the `./dist` bind mount with the catalog build of `grafana-pyroscope-app`, and installs `grafana-llm-app` directly in `Dockerfile.remote` (since disabling preinstall also disables `GF_INSTALL_PLUGINS`).
- Enables the span profiles heatmap feature toggles via a configurable `GF_FEATURE_TOGGLES_ENABLE` env var and makes the remote Grafana port overridable with `GRAFANA_PORT`.

## Test 🧪

- [x] Ran `server:remote` with `REMOTE_TRACE_*` credentials set and confirmed the "Dev (tempo-dev-01)" Tempo datasource is provisioned and selectable in the span exemplar table
- [x] Verified the mounted `./dist` build of `grafana-pyroscope-app` is used (not overwritten) with `GF_PLUGINS_PREINSTALL_DISABLED=true` on Grafana 12.3+
- [x] Confirmed `grafana-llm-app` is installed in the remote image and the profiles heatmap renders with the enabled feature toggles

### DIFF
--- a/Dockerfile.remote
+++ b/Dockerfile.remote
@@ -3,13 +3,6 @@ ARG grafana_version=12.3.0@sha256:96a793a92c9a77cf543d6e5c55100cd296ed9e22487dc3
 
 FROM grafana/${grafana_image}:${grafana_version}
 
-ARG backendUrl
-ARG basicAuthUser
-ARG basicAuthPassword
-ARG traceBackendUrl
-ARG traceBasicAuthUser
-ARG traceBasicAuthPassword
-
 # Make it as simple as possible to access the grafana instance for development purposes
 # Do NOT enable these settings in a public facing / production grafana instance
 ENV GF_AUTH_ANONYMOUS_ORG_ROLE="Admin"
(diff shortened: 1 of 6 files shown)


Dockerfile.remote
```diff
@@ -20,8 +23,19 @@ ENV REMOTE_BACKEND_URL=$backendUrl
 ENV REMOTE_BASIC_AUTH_USER=$basicAuthUser
 ENV REMOTE_BASIC_AUTH_PASSWORD=$basicAuthPassword
 
+# Credentials for the remote Tempo datasource (span profiles -> traces)
+ENV REMOTE_TRACE_BACKEND_URL=$traceBackendUrl
+ENV REMOTE_TRACE_BASIC_AUTH_USER=$traceBasicAuthUser
+ENV REMOTE_TRACE_BASIC_AUTH_PASSWORD=$traceBasicAuthPassword
+
 USER root
```
simonswine, line 14: We should not build secrets into the dockerfile, they can be accidentally leaked when the underlying docker image gets pushed.

(This is already happening for the current credentials)

I have a stacked PR #1058 which should fix this

simonswine (commented):
Thanks for following up, I think we should fix the secrets.

simonswine (approved):
LGTM, but my approval doesn't count anymore

matyax (approved):

PR #1057, opened 2026-07-22
fix: show persistent settings-fetch error only on settings page

simonswine:
We currently show this not really actionable error toasts:

image

They mostly happens for OSS users, that point the datasource directly at the query-frontend rather than at a proxy that handles tenant-setttings.

I have decided to improve on this and only show those errors as part of the error page:

image


L2D2Grafana (dismissed):
Image So nice not to get spammed by plugin setting failures!

simonswine:
Rebase and lost approval 😢

matyax (approved):

PR #1058, opened 2026-07-22
fix: avoid baking remote backend secrets into the Dockerfile

simonswine:
Improve on #1054.

Previously `REMOTE_BACKEND_URL`, `REMOTE_BASIC_AUTH_USER`, `REMOTE_BASIC_AUTH_PASSWORD`
and the equivalent `REMOTE_TRACE_*` credentials for the Tempo datasource were passed as
Docker build args and baked into the image via `ENV` instructions plus `sed`-substituted
provisioning files. That persists secrets in image layers/history.


L2D2Grafana (approved):
👏

PR #1065, opened 2026-07-24
fix(header): use ButtonGroup for plugin header toolbar buttons

L2D2Grafana:
✨Description

The group of buttons at the far right of the header (upload, share, settings, plugin info, and the recording-rules dropdown when enabled) was wrapped in an `appMiscButtons` div that drew its own border and background, then stripped the borders off the buttons inside. The result did not match the standard secondary/canvas button styling used by the neighbouring time-range and refresh controls.

This replaces the custom wrapper with `ButtonGroup` from `@grafana/ui`, which is the standard component for this. No custom styling is needed here.

Same fix as [grafana/logs-drilldown#2009](https://github.com/grafana/logs-drilldown/pull/2009); originally reported in [Slack](https://raintank-corp.slack.com/archives/C075BDBTX96/p1784800446330289).

📖 Summary of the changes

- `src/shared/ui/PluginHeaderToolbar.tsx`: the `appMiscButtons` wrapper div is now a `ButtonGroup`.
- Removed the `appMiscButtons` style block. It set a container border + `background.secondary` and then applied `button { border: none; border-radius: 0; }` (plus the same on `:hover`/`:focus`) to every child, which is what removed the standard button styling.
- `ButtonGroup` handles joining the buttons (unsets the inner border radii and adds the divider) while each button keeps its own border and background.
- All five buttons in the group were already `ToolbarButton variant="canvas"`, so no change was needed there.

Net: 3 insertions, 24 deletions.

🧪 How to test?
Screenshot 2026-07-24 at 2 18 58 PM

### DIFF
--- a/src/shared/ui/PluginHeaderToolbar.tsx
+++ b/src/shared/ui/PluginHeaderToolbar.tsx
@@ -2,7 +2,7 @@ import { css, cx } from '@emotion/css';
 import { GrafanaTheme2 } from '@grafana/data';
 import { t } from '@grafana/i18n';
 import { useChromeHeaderHeight, usePluginComponent } from '@grafana/runtime';
-import { Dropdown, ErrorBoundary, Field, Icon, Menu, ToolbarButton, useStyles2 } from '@grafana/ui';
+import { ButtonGroup, Dropdown, ErrorBoundary, Field, Icon, Menu, ToolbarButton, useStyles2 } from '@grafana/ui';
 import { SaveSearchButton } from '@shared/components/SavedSearches/SaveSearchButton';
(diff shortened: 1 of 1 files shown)


simonswine (approved):
Not having a particular good UI eye, but this LGTM.

joey-grafana (approved):

joey-grafana:
Thanks Liza! Is there a PR for Metrics/Traces? I can create those if not :)

L2D2Grafana:
> Thanks Liza! Is there a PR for Metrics/Traces? I can create those if not :)

I can update Traces, Metrics had the fix already. It's not a noticeable for traces yet
 
Screenshot 2026-07-27 at 7 11 50 AM

PR #1068, opened 2026-07-27
fix: time picker menu position

### LINKED ISSUE #1016: [bug] Picker hidden under the side menu
### Bug description

- **Grafana version:** v13.1.0-27262365585
- **Plugin version:** v2.0.7+d6b2ddadf5563e3c8631dbd61b2951cda11e5a35

The overlay between the time picker and the side menu is broken - the picker hides behind the side menu.
If the screen width is too narrow and the picker gets pushed to the left, upon opening it the side menu overlays it, cutting it in half. 
I'm providing a screenshot of the issue.

This is happening in cloud. This seems to happen across all of the Drilldown products; I've checked other entities and they seem to be behaving correctly (picker has more priority than the side menu).

### Steps to reproduce

1. Go to Drilldown - Profiles
2. Zoom in if needed to reduce the width and make sure that the picker is positioned closer to the left
3. Open it - the bug should manifest itself

Image

#### Expected behaviour

The picker should show over the side menu.

joey-grafana:
### ✨ Description

Fixes: https://github.com/grafana/profiles-drilldown/issues/1016

When the time picker sits near the left edge, its panel opens with right: 0 and grows under the docked mega menu. Align the overlay with left: 0 so it opens into the page instead.

### DIFF
--- a/src/shared/ui/PluginHeaderToolbar.tsx
+++ b/src/shared/ui/PluginHeaderToolbar.tsx
@@ -213,6 +213,12 @@ const getStyles = (theme: GrafanaTheme2, chromeHeaderHeight: number, isEmbedded:
     display: flex;
     align-items: center;
     gap: ${theme.spacing(1)};
+
+    /* TimeRangePicker defaults to right:0; open into the page so the docked nav does not clip it */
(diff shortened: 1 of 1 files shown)


simonswine (approved):
LGTM

Issue #1053, opened 2026-07-17
Improve span selector placement in profiles exploration


simonswine:
Follow up from https://github.com/grafana/profiles-drilldown/pull/934.

Move or duplicate the span ID selector into the top query/filter experience so users can see and select the active span without returning to the flame graph control.

This needs to account for the existing profile ID selector, which is currently presented alongside the span selector, and define a cohesive top-level presentation for both selectors.

There is also an upcoming trace ID selector coming.

PR #1077, opened 2026-07-30
feat(flame graph): add a span ID selector

L2D2Grafana:
### ✨ Description

Issue: #1053 (Improve span selector placement in profiles exploration)

Updated the Span Id to use combobox and show span ID with the caveat that it is not an exhaustive list. If people think this treatment of the Span ID is better I can update the Profile ID too.
Screenshot 2026-08-10 at 1 24 00 PM

https://github.com/user-attachments/assets/6a584798-9a10-4f3c-8a1f-6dc8cc636727

### DIFF
--- a/src/locales/en-US/grafana-pyroscope-app.json
+++ b/src/locales/en-US/grafana-pyroscope-app.json
@@ -676,7 +676,7 @@
     }
   },
   "span-id-filter": {
-    "label": "Span ID",
+    "exemplar-group": "Sampled span exemplars (not an exhaustive list)",
     "no-options": "No span profiles in the current time range",
     "placeholder": "Filter by span ID"
   },
--- a/src/locales/en-US/grafana-pyroscope-app.json
+++ b/src/locales/en-US/grafana-pyroscope-app.json
@@ -675,6 +675,11 @@
       }
     }
   },
+  "span-id-filter": {
+    "label": "Span ID",
+    "no-options": "No span profiles in the current time range",
+    "placeholder": "Filter by span ID"
+  },
   "timeseries": {
     "menu": {
       "add-to-dashboard": "Add to dashboard",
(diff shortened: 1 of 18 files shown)


simonswine (commented):
I have some reservations about this approach. The main sticking points for me are:

- The heatmap only exposes a sampled subset of span IDs, typically one exemplar per bucket, while many more spans may exist. Using those exemplars as dropdown options could give users the impression that the list is exhaustive. We currently do not have an API for discovering all span IDs, and building one would likely require prefix search and pagination.

- Placing the selector in the top filter row visually associates it with the timeline. However, the timeline shows full process profiles rather than span-scoped profiles, so it is not particularly meaningful once a span is selected. It may also imply that the span filter applies consistently to both the timeline and the flame graph when it does not.
If the primary use case is pasting a known span ID, I wonder whether a free-form input closer to the flame graph would communicate the scope more accurately than a top-level picker populated from heatmap exemplars.

L2D2Grafana:
> I have some reservations about this approach. The main sticking points for me are:
> 
> * The heatmap only exposes a sampled subset of span IDs, typically one exemplar per bucket, while many more spans may exist. Using those exemplars as dropdown options could give users the impression that the list is exhaustive. We currently do not have an API for discovering all span IDs, and building one would likely require prefix search and pagination.
> * Placing the selector in the top filter row visually associates it with the timeline. However, the timeline shows full process profiles rather than span-scoped profiles, so it is not particularly meaningful once a span is selected. It may also imply that the span filter applies consistently to both the timeline and the flame graph when it does not.
>   If the primary use case is pasting a known span ID, I wonder whether a free-form input closer to the flame graph would communicate the scope more accurately than a top-level picker populated from heatmap exemplars.

@simonswine that makes sense! Is there another way you'd like to see it implemented like perhaps only appearing in the header once user selects a span id and only display the one id kind of like what it does now, only in the header? Or should I close this for now because we don't know 🤔

simonswine:
@L2D2Grafana I think a free form entry might make sense (explore has that), but I doubt many people are using it, as it is quite hard to get that span ID. More sense might be a removal only thing a bit like below.  

But for that the timeline concern is still valid.

L2D2Grafana:
> @L2D2Grafana I think a free form entry might make sense (explore has that), but I doubt many people are using it, as it is quite hard to get that span ID. More sense might be a removal only thing a bit like below.
> 
> But for that the timeline concern is still valid.

Okay updated Span ID to only show on the heatmap panel.  It behaves much the same as the original Span ID but now shows options and uses the canonical grafana/ui combobox. If we like this I can take a 🔪 stab at profile ID too.

L2D2Grafana:
I think this is not something we want right now.

PR #1088, opened 2026-08-05
feat: copy gcx pprof commands

simonswine:
I find myself very often wanting to investigate a particular flamegraph further locally. This adds a copy GCX command to clipboard button to the single and the diff view.

Let me know how you feel about this

image

image

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx
@@ -4,9 +4,12 @@ import { t, Trans } from '@grafana/i18n';
 import { SceneComponentProps, sceneGraph, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
 import { Spinner, useStyles2 } from '@grafana/ui';
 import { FlameGraph } from '@shared/components/FlameGraph/FlameGraph';
+import { displayError, displaySuccess } from '@shared/domain/displayStatus';
 import { reportInteraction } from '@shared/domain/reportInteraction';
+import { useMaxNodesFromUrl } from '@shared/domain/url-params/useMaxNodesFromUrl';
 import { useToggleSidePanel } from '@shared/domain/useToggleSidePanel';
 import { getProfileMetric, ProfileMetricId } from '@shared/infrastructure/profile-metrics/getProfileMetric';
+import { DEFAULT_SETTINGS } from '@shared/infrastructure/settings/PluginSettings';
 import { useFetchPluginSettings } from '@shared/infrastructure/settings/useFetchPluginSettings';
 import { DomainHookReturnValue } from '@shared/types/DomainHookReturnValue';
 import { FlamebearerProfile } from '@shared/types/FlamebearerProfile';
@@ -15,6 +18,8 @@ import { Panel } from '@shared/ui/Panel/Panel';
 import { PyroscopeLogo } from '@shared/ui/PyroscopeLogo';
 import React, { useEffect, useMemo } from 'react';
 
+import { buildGcxPprofCommand } from '../../../../domain/buildGcxPprofCommand';
+import { getPprofExportFilename } from '../../../../domain/getPprofExportFilename';
 import { useBuildPyroscopeQuery } from '../../../../domain/useBuildPyroscopeQuery';
 import { useGrafanaAssistant } from '../../../../domain/useGrafanaAssistant';
 import { ProfilesDataSourceVariable } from '../../../../domain/variables/ProfilesDataSourceVariable';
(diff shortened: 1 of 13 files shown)


src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneExportMenu/SceneExportMenu.tsx
```diff
@@ -219,6 +247,15 @@ export class SceneExportMenu extends SceneObjectBase<SceneExportMenuState> {
             />
             <Menu.Item label={t('export-menu.json', 'json')} onClick={actions.downloadJson} />
             <Menu.Item label={t('export-menu.pprof', 'pprof')} onClick={actions.downloadPprof} />
+            <Tooltip
+              content={t('export-menu.gcx-command-tooltip', 'Copy the gcx command to download this profile as pprof')}
+            >
+              <Menu.Item
+                icon="clipboard-alt"
```
aleks-p: nit, should this be the copy icon?

aleks-p (commented):
This is a great idea!

I left one nit. I haven't looked at the code itself.

joey-grafana:
Nice! It would prob be tidier to create two sections/groups in the menu, wdyt? As the copy icon and the gpx command option seems a bit out of place just at the bottom and is a lot longer.

simonswine:
After the two changes it looks like this now:

image

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneExportMenu/domain/getExportFilename.ts
```diff
@@ -1,8 +1 @@
-import { TimeRange } from '@grafana/data';
-import { parseQuery } from '@shared/domain/url-params/parseQuery';
-
-export function getExportFilename(query: string, timeRange: TimeRange) {
-  const { serviceId, profileMetricId } = parseQuery(query);
-  const dateString = `${timeRange.from.format('YYYY-MM-DD_HHmm')}-to-${timeRange.to.format('YYYY-MM-DD_HHmm')}`;
-  return `${serviceId.replace(/\//g, '-')}_${profileMetricId}_${dateString}`;
-}
+export { getPprofExportFilename as getExportFilename } from '../../../../../domain/getPprofExportFilename';
```
L2D2Grafana, line 1: This is a bit weird. There is one function with two different names. I'd keep the original name and import it from the new place it was moved.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/components/SceneExportMenu/SceneExportMenu.tsx
```diff
@@ -153,6 +156,30 @@ export class SceneExportMenu extends SceneObjectBase<SceneExportMenuState> {
...
 
+    const copyGcxCommand = async () => {
+      const filename = `${getExportFilename(query, timeRange)}.pb.gz`;
+      const command = buildGcxPprofCommand({
+        dataSourceUid,
+        query,
+        timeRange,
+        maxNodes: maxNodes || DEFAULT_SETTINGS.maxNodes,
+        filename,
+        profileIds: profileIdSelector ? [profileIdSelector] : undefined,
```
L2D2Grafana, line 167: `profileIdSelector`/`spanSelector` never actually reach this code. `SceneFlameGraph` passes them into `<Component />`, but `static Component` (line 231) destructures only `{ model, query, timeRange }` and calls `useSceneExportMenu({ query, timeRange })`, so both selectors are always `undefined` here — the copied command never includes `--profile-id`/`--span-id`. TypeScript doesn't flag it because both props are optional.

```tsx
static Component = ({ model, query, timeRange, profileIdSelector, spanSelector }: SceneComponentProps<SceneExportMenu> & ExtraProps) => {
  const { data, actions } = model.useSceneExportMenu({ query, timeRange, profileIdSelector, spanSelector });
```

src/pages/ProfilesExplorerView/components/SceneExploreDiffFlameGraph/components/SceneDiffFlameGraph/SceneDiffFlameGraph.tsx
```diff
@@ -101,6 +107,35 @@ export class SceneDiffFlameGraph extends SceneObjectBase<SceneDiffFlameGraphStat
     );
     const hasMissingSelections = !isDiffQueryEnabled;
 
+    const copyGcxCommands = async () => {
```
L2D2Grafana, line 110: This and `copyGcxCommand` in `SceneExportMenu` duplicate the clipboard-write `reportInteraction`, success/error-toast tail. This could be extracted.

L2D2Grafana (approved):
Cool idea, maybe we should do this in logs drilldown. Left some minor comments.

Image

joey-grafana (approved):

PR #1121, opened 2026-09-10
feat(flame-graph): progressive loading with call-site refinement on focus

aleks-p:
Tenants raise `max_nodes` to 1M so they never see an `other` node, and pay for it in latency: on a dev cell the same 15 service/range combinations took 3.4x longer to first paint at the median (11.5x worst case), shipped 2-46 MB instead of 0.7-0.9 MB, and 4 of 15 windows failed outright at the datasource's 30s timeout.

This queries a bounded tree (16384 nodes) and then refines only the truncated subtrees the user actually focuses, using the `stackTraceSelector` call site added in grafana/grafana-pyroscope-datasource#57. Configured `max_nodes` is ignored while it is on. Total backend CPU for a whole session, refinements included, is 0.8x a single 1M query.

Enabled by default via `progressiveFlamegraphs` in plugin settings.

_Video: TODO_

Notes:
- Draft: needs `onFocusChange` and `loadingPaths` from grafana/grafana#132316 in a released `@grafana/flamegraph`. There is a TODO at the import with the local-build workaround.

### DIFF
--- a/src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
+++ b/src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
@@ -1,8 +1,12 @@
 import { css } from '@emotion/css';
-import { createTheme, GrafanaTheme2, LoadingState, TimeRange } from '@grafana/data';
+import { createTheme, DataFrame, GrafanaTheme2, LoadingState, TimeRange } from '@grafana/data';
+// TODO: onFocusChange and loadingPaths are not in a released @grafana/flamegraph yet (see grafana/grafana#132316).
+// Until it ships, point the dependency at a local build:
+//   "@grafana/flamegraph": "file:./grafana-flamegraph-local.tgz"
 import { FlameGraph, Props as FlameGraphProps } from '@grafana/flamegraph';
 import { t, Trans } from '@grafana/i18n';
 import {
+  sceneGraph,
   SceneComponentProps,
   SceneObjectBase,
   SceneObjectState,
@@ -14,6 +18,7 @@ import { useMaxNodesFromUrl } from '@shared/domain/url-params/useMaxNodesFromUrl
 import { useToggleSidePanel } from '@shared/domain/useToggleSidePanel';
 import { useFlagMetricsFromProfiles } from '@shared/infrastructure/featureFlags/featureFlags';
 import { getProfileMetric, ProfileMetricId } from '@shared/infrastructure/profile-metrics/getProfileMetric';
+import { DEFAULT_SETTINGS } from '@shared/infrastructure/settings/PluginSettings';
 import { useFetchPluginSettings } from '@shared/infrastructure/settings/useFetchPluginSettings';
 import { DomainHookReturnValue } from '@shared/types/DomainHookReturnValue';
 import { InlineBanner } from '@shared/ui/InlineBanner';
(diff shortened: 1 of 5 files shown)


L2D2Grafana (commented):
Wow this is some advanced stuffs 👏 . It's not many ui developers who'd try to write an observable from scratch. It's cool, but it can be hard to maintain in the future. This is usually where we reach for a library. The best bet would be to use RXjs. It's already a dependency etc. The current controller manually implements an observable workflow using queue, pump, active, inFlight, cancelled, and emit, but cancel() does not unsubscribe from requests already in flight. RxJS already returns an Observable. I'd try updating concurrency, cancellation, state accumulation, and error handling as an RxJS pipeline.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
```diff
@@ -1,9 +1,14 @@
 import { css } from '@emotion/css';
-import { createTheme, GrafanaTheme2, LoadingState, TimeRange } from '@grafana/data';
+import { createTheme, DataFrame, GrafanaTheme2, LoadingState, TimeRange } from '@grafana/data';
+// TODO: onVisibleTruncatedPathsChange and loadingPaths are not in a released @grafana/flamegraph yet
+// (see grafana/grafana#132316).
+// Until it ships, point the dependency at a local build:
+//   "@grafana/flamegraph": "file:./grafana-flamegraph-local.tgz"
```
L2D2Grafana, line 6: This could also be fun to try
```
Then build once and link:
cd packages/grafana-flamegraph
pnpm install --frozen-lockfile=false   # or whatever this workspace needs
pnpm run build
pnpm link --global

cd code/plugins/profiles-drilldown
pnpm link --global @grafana/flamegraph
```

src/pages/ProfilesExplorerView/infrastructure/flame-graph/progressiveProfileTree.ts
```diff
@@ -0,0 +1,125 @@
...
+export function treeToDataFrame(root: ProfileTreeNode, unit: string): DataFrame {
+  const labelValues: string[] = [];
+  const levelValues: number[] = [];
+  const selfValues: number[] = [];
+  const valueValues: number[] = [];
+
+  const stack: Array<{ node: ProfileTreeNode; level: number }> = [{ node: root, level: 0 }];
+
+  while (stack.length) {
+    const { node, level } = stack.shift()!;
```
L2D2Grafana, line 67: Fun javascript fact shift() re-indexes every element  O(n), if you can get away with .pop() or this 
```
while (head < queue.length) {
  const item = queue[head++]; // O(1)
  queue.push(...nextItems);
}
```
It can speed things up.

src/pages/ProfilesExplorerView/components/SceneExploreServiceFlameGraph/SceneFlameGraph.tsx
```diff
@@ -47,6 +60,8 @@ interface SceneFlameGraphState extends SceneObjectState {
   $timeRange?: SceneTimeRange;
   $data: SceneQueryRunner;
   lastTimeRange?: TimeRange;
+  progressiveFrame?: DataFrame;
+  progressiveProgress?: ProgressiveProgress;
```
L2D2Grafana, line 64: Some of the progressive names get a bit obtuse. We call a lot of this stuff progressive loading or lazy loading.

aleks-p:
Thank you very much for the review @L2D2Grafana! 

I've discovered a few issues with my original approach, especially around how the Call Tree view works progressive (maybe I just rename that to lazy) loading. I plan to rework this, but will make sure to incorporate your feedback in the next iteration. 

I'll close the draft PRs for now and let you know when I have something ready.

PR #1130, opened 2026-09-15
fix: move Profile types and Labels into the panel kebab menu

### LINKED ISSUE #1129: [BUG]: Panel actions truncate namespaced service names
## Describe the bug

Each profile panel title includes three inline actions: **Profile types**, **Labels**, and **Flame graph**. These actions consume enough horizontal space that namespaced service names are truncated too aggressively.

For many panels, the service name is rendered as something like `alertmanager/...`, which hides the meaningful part of the name and makes panels difficult to distinguish.

## To Reproduce

1. Open the Profiles Drilldown services view.
2. Use a grid layout with namespaced service names.
3. Observe the panel titles.
4. Notice that the inline actions cause names such as `alertmanager/...` to be ellipsized.

## Expected behavior

Service names should remain readable enough to distinguish panels, especially the namespace and service suffix. Panel actions should not take priority over the identity of the service.

## Possible fix

Move **Profile types**, **Labels**, and **Flame graph** into the panel title ellipsis menu. This would preserve access to the actions while giving the service name the available title width.

An alternative would be to retain the inline actions but use a responsive layout or show the full service name in a tooltip, while prioritizing more of the service name over the action labels.

## Environment

- Area: Profiles Drilldown services grid
- View: All services
- Profile type: `process_cpu/cpu`

## Screenshot

Image

## AI disclosure (optional)

- [x] This issue was substantially generated with AI assistance.

joey-grafana:
### ✨ Description

Fixes: https://github.com/grafana/profiles-drilldown/issues/1129
fix: move Profile types and Labels into the panel kebab menu

- Move **Profile types** and **Labels** from the timeseries panel header into the kebab menu (above Open in Explore) so namespaced service names are not truncated to `alertmanager/...`.
- Keep **Flame graph** as an icon-only header action (same hit target as Favorite) and always show the panel menu.
- Timeseries panels relocate those two actions automatically; views still declare them in `headerActions`.

Before

Screenshot 2026-09-15 at 09 48 15

After

Screenshot 2026-09-15 at 09 42 58


hugohaggmark (dismissed):
LGTM 🎉  and works as expected, do we currently have tracking on the existing labels? Would be interesting to see how much this change will impact

hugohaggmark (approved):

PR #1135, opened 2026-09-18
chore: add pre-commit hook

hugohaggmark:
**What is this feature?**

Adds a Lefthook pre-commit hook that runs Prettier on staged JavaScript and TypeScript files, plus formats the existing repository with Prettier.

**Why do we need this feature?**

So formatting stays consistent before commits.

**Which issue(s) does this PR fix?**:

Fixes #

**Special notes for your reviewer:**

The formatting-only commit is intentionally separate from the Lefthook setup.


L2D2Grafana (commented):
This is a cool idea! Logs-drilldown uses husky. We could be persuaded to use lefthook. You can checkout our lint-staged it also enforces eslint. The only thing I'd change here is maybe not run it on the translation files, it could cause a lot of churn as they are automatically updated.

joey-grafana (dismissed):
Agreed, we should ignore translation files but otherwise this LGTM 👍

hugohaggmark:
@L2D2Grafana @joey-grafana the `lefthook.yml` doesn't include any `*.json` files right? or am I missing something?

```yml
pre-commit:
  parallel: true
  commands:
    frontend-lint:
      glob: '*.{js,jsx,ts,tsx,mjs,cjs,mts,cts}'
      run: |
        pnpm prettier --write {staged_files}

      stage_fixed: true
```

.release-please-manifest.json
```diff
@@ -1 +1 @@
-{".":"2.3.1"}
+{ ".": "2.3.0" }
```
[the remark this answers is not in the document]
hugohaggmark: oops my bad 🙈 , I must have merged it incorrectly, i'll fix 👍

pnpm-lock.yaml
```diff
@@ -39,7 +39,6 @@ overrides:
   react-router@^6: ^6.30.6
```
joey-grafana, line 39: pnpm prettier --write --list-different . reformatted all of pnpm-lock.yaml, and it still isn’t in .prettierignore. format:fix / lint:fix and the next pnpm lockfile rewrite will keep producing thousand-line formatting diffs on top of the real lefthook entry.

Add pnpm-lock.yaml to .prettierignore and restore lockfile formatting to what pnpm emits, leaving only the lefthook package lines.

lefthook.yml
```diff
@@ -0,0 +1,9 @@
+pre-commit:
```
joey-grafana, line 1: lefthook has a postinstall and isn’t allowlisted

lefthook@2.1.14 runs node postinstall.js (lefthook install). This repo uses strictDepBuilds: true and an allowBuilds list that doesn’t include lefthook. Fresh pnpm install / CI can fail on that script even though hooks are documented as opt-in (pnpm lefthook install).

Keep hooks manual: add lefthook to ignoredBuiltDependencies (or the pnpm 11 equivalent). Don’t put it in allowBuilds unless you actually want postinstall to auto-install git hooks.

.release-please-manifest.json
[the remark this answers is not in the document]
hugohaggmark: fixed in https://github.com/grafana/profiles-drilldown/pull/1135/commits/94808beb209e41aa247b86ba0eb78c018ac4cf58 👍

CONTRIBUTING.md
```diff
@@ -117,6 +117,22 @@ In case of doubt, have a look at ["Profiles Drilldown frontend architecture"](do
 
 We use [ESLint](https://eslint.org/) and [Prettier](https://prettier.io/) to lint and format our code. These will be run in a pre-commit hook, but you can also setup your IDE to run them on save.
 
+#### Configure precommit hooks
+
+We use pre-commit hooks (via [lefthook](https://github.com/evilmartians/lefthook)) to lint, fix, and format code as you commit your changes.
```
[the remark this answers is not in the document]
hugohaggmark: https://github.com/grafana/profiles-drilldown/pull/1135/commits/94808beb209e41aa247b86ba0eb78c018ac4cf58

pnpm-lock.yaml
hugohaggmark: fixed in https://github.com/grafana/profiles-drilldown/pull/1135/commits/7b7a9efe8e4007d51fbf1b4a044e3b2910f49224 👍

lefthook.yml
hugohaggmark: great catch 🙌 , looking into it now 👍

hugohaggmark: fixed in https://github.com/grafana/profiles-drilldown/pull/1135/commits/c753763d1b6359ff5220922ccd602b0416f35631 👍

pnpm-workspace.yaml
```diff
@@ -61,3 +61,5 @@ overrides:
 minimumReleaseAgeExclude:
   # Renovate security update: pnpm@11.11.0
   - pnpm@11.11.0
+onlyBuiltDependencies:
```
joey-grafana: Same issue as https://github.com/grafana/traces-drilldown/pull/913#pullrequestreview-5287958880

But honestly I feel like it's better to keep it out of allowBuilds and run pnpm lefthook install by hand, which matches this repo’s ignoreScripts policy and CONTRIBUTING. A compromised lefthook release with allowBuilds would execute on install and could rewrite git hooks; doing it explicitly keeps that on the contributor, not on every install.

hugohaggmark: thanks for the detailed explaination 🙏 , I think I understand now 🙈 It was never my intention to run it with every install but to have it as a opt in / opt out kind of thing as the README describes. So I'll remove this and close https://github.com/grafana/traces-drilldown/pull/913

joey-grafana (approved):

